PR #59, opened 2026-04-20
RES-125: `_` type holes as inference placeholders

EricSpencer00:
## Summary

- Parser now accepts `_` in all type annotation positions (`let x: _`, `fn foo(_ x) -> _`, return types, struct fields).
- Typechecker maps `"_"` to `Type::Any` so existing compatibility checks accept it without misidentifying it as a struct type named `_`.
- HM inferer (`infer.rs`) mints a fresh `Type::Var` per `_` annotation — each hole is independently resolved. This works naturally via the existing `parse_primitive_type() -> None` fallback path that already calls `self.fresh()`.
- Added `Inferer::env_type()` public accessor to support cross-module tests.
- Fixed pre-existing build breakage in this worktree (RES-228 in-progress): added `notify` dependency, `imports::collect_watched_paths` stub, and missing `InvariantStatement` / `StringInterpolation` match arms in `free_vars.rs`.

## Test plan

- [ ] `res125_underscore_let_annotation_parses` — `let x: _ = 42` parses with `type_annot = "_"`
- [ ] `res125_underscore_fn_param_annotation_parses` — `fn foo(_ x) -> int` parses with param type `"_"`
- [ ] `res125_underscore_fn_return_annotation_parses` — `fn bar(int x) -> _` parses with return type `"_"`
- [ ] `res125_each_underscore_becomes_independent_inference_var` (feature=infer) — two `_ ` params get distinct `Type::Var` ids
- [ ] `res125_let_underscore_infers_int_from_arithmetic` (feature=infer) — `let x: _ = 3 + 2` infers `x: Int`
- [ ] All 742+ existing tests continue to pass (no regressions)


EricSpencer00:
Conflicts with main. Type holes (RES-125) partially landed in #78. The remaining work should be a fresh PR on current main.

PR #71, opened 2026-04-20
RES-246: fix LSP walk tests to use .res extension after rename

EricSpencer00:
## Summary

- `walk_resilient_files` was updated to match `*.res` files (not `*.rs`), but three tests that seed temp files were not updated, causing them to silently find nothing and fail their assertions.
- Rename scaffold temp files from `*.rs` to `*.res` in the two unit tests specified by the ticket, plus one additional integration smoke test with the same root cause.

## Test changes

These are bug-fix-only test updates — no assertions are weakened or removed.

| File | Test | Change |
|---|---|---|
| `resilient/src/lsp_server.rs` | `walk_resilient_files_finds_rs_files_recursively` | Scaffold files `a.rs`, `b.rs`, `c.rs`, `d.rs` → `*.res`; assertions updated to match |
| `resilient/src/lsp_server.rs` | `workspace_index_spans_multiple_files` | Scaffold files `mod_a.rs`, `mod_b.rs` → `*.res`; URI suffix assertions updated |
| `resilient/tests/lsp_smoke.rs` | `lsp_workspace_symbol_searches_multiple_files` | Scaffold files `mod_a.rs`, `mod_b.rs` → `*.res`; filename assertions updated |

The third test (`lsp_workspace_symbol_searches_multiple_files`) was already failing before this PR for the same reason — included here since it is the identical fix.

## Verification

- `cargo test --manifest-path resilient/Cargo.toml --features lsp` — 791 unit tests + all integration tests pass (0 failures)
- `cargo clippy --manifest-path resilient/Cargo.toml --all-targets --features lsp -- -D warnings` — clean


EricSpencer00:
Conflicts with main. LSP test .res extension fix should be re-applied as a fresh PR on current main.

PR #80, opened 2026-04-21
RES-284 RES-285: Value::Map expect fix + VM assert/assume support

EricSpencer00:
## Summary

- **RES-284** (P4): Replace `.expect("key is from map")` with idiomatic `&m[k]` in `impl Display for Value::Map` (`resilient/src/main.rs`). The invariant is sound (keys are collected from `m.keys()`), so the index form is the idiomatic Rust choice. No behaviour change.

- **RES-285** (P3): Add lowering for `assert()` and `assume()` in the bytecode VM compiler (`--vm` path). Previously both fell through to `CompileError::Unsupported`, silently dropping safety-critical runtime checks.

  Implementation:
  - New `Op::RuntimeError` opcode in `bytecode.rs` — pops a message string and halts with `VmError::AssertionFailed(String)`.
  - VM dispatch arm in `vm.rs`.
  - Disasm arm in `disasm.rs`.
  - `compile_assert_or_assume()` helper in `compiler.rs` lowers both nodes to: compile condition → `JumpIfTrue(skip)` → push message const → `RuntimeError` → patch skip.
  - Arms added to both `compile_stmt` (top-level) and `compile_stmt_in_fn` (fn bodies).
  - 5 new tests: assert true passes, assert false fails with "assertion failed", assume true passes, assume false fails, assert inside fn body works.

  Note: the JIT path (`--jit`) has the same gap as a pre-existing issue; out of scope for this PR.

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml` — 724 tests, 0 failures
- [x] `cargo fmt --manifest-path resilient/Cargo.toml --check` — clean

## Test changes

None — all new tests, no existing tests modified.

### DIFF
--- a/.board/designs/concurrency.md
+++ b/.board/designs/concurrency.md
@@ -1,0 +1,190 @@
+# Concurrency Design Note — RES-208
+
+**Status:** Parse-scaffolding landed; scheduler not yet wired.
+**Feature flag:** `concurrency-preview`
+**Related tickets:** RES-191 (`@pure`), RES-192 (`@io`), RES-193 (effect polymorphism), RES-208 (this note)
+
+---
+
+## Actor Semantics
+
+### State Ownership
+
+Each actor owns its state exclusively. No reference to an actor's
+internal state escapes the actor's body. This is the
+data-race-freedom property stated positively: there is no shared
+mutable state to race on.
+
+In the preview syntax:
+
+```
+actor Counter {
+    let mut count = 0;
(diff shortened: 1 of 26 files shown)


EricSpencer00:
Conflicts with main after bytecode/compiler/vm changes from FFI v2 and TCO. The Value::Map fix and VM assert/assume support should be re-applied as a fresh PR.

Issue #267, opened 2026-04-26
Tree-walker stack-overflows at ~1000 deep recursion


EricSpencer00:
## Summary

A simple linear-recursive function blows the host process's stack at depth
1000. For comparison, CPython handles ~10,000 by default and Rust ~100,000.
The tree-walker's per-call frame is too heavy for safety-critical workloads
that the docs frame as the headline target.

## Reproduction

```rz
fn count(int n) {
    if n <= 0 { return 0; }
    return 1 + count(n - 1);
}
fn main() { println(count(1000)); return 0; }
main();
```

```text
$ rz deep.rz

thread 'main' (49930870) has overflowed its stack
fatal runtime error: stack overflow, aborting
```

`count(500)` works; `count(1000)` doesn't.

## Why this matters

Embedded targets typically have constrained stacks, but the tree-walker is
the **host-side** interpreter — even with the OS giving us ~8 MB of stack,
~1000 deep recursion exhausts it. That implies each frame is ~8 KB, which is
heavy for a tree walker on the value layer the runtime crate already keeps
alloc-free.

## Suggested directions (any one is enough)

- Trampoline the tree-walker — convert the recursive `eval` over `Node` into
  an explicit work stack on the heap, so call depth is bounded by `Vec`
  capacity rather than the host stack.
- Tail-call optimization where applicable (`return f(...)` with no work
  after) — straight TCO covers the cheap cases.
- Bounded recursion as a typed runtime error rather than `abort()` so a
  `live` block can catch it.

## Severity

Medium. Most realistic Resilient programs don't recurse 1000 deep — the
language doesn't even have a standard collection type that needs deep
recursion. But "the runtime crashes the host" is a worse failure mode than
"the runtime returns an error" given the project's safety-critical framing.

EricSpencer00:
Claiming — minimal frame-size fix + stack bump.

PR #290, opened 2026-04-27
RES-319: newtype pattern

### LINKED ISSUE #111: RES-319: Newtype pattern — `newtype Meters = Float` for type-safe units
## Summary

Type aliases (RES-88) make names transparent. For safety-critical unit systems (meters, volts, seconds) the compiler should reject mixing incompatible units even if the underlying type is the same. `newtype` creates a distinct type that does not unify with its base.

\`\`\`
newtype Meters = Float;
newtype Seconds = Float;

fn speed(d: Meters, t: Seconds) -> Float { return d / t; }

// compile error: cannot pass Seconds where Meters expected
let bad = speed(10.0s, 5.0m);
\`\`\`

## Acceptance criteria

- `newtype Name = BaseType;` declares a new, non-interchangeable type.
- Construction via `Name(value)` converts a base-type value to the newtype.
- Arithmetic operators are inherited from the base type and return the newtype.
- Mixing two different newtypes of the same base in arithmetic is a type error.
- Golden test: a `Meters`/`Seconds`/`MetersPerSecond` example that correctly rejects a unit error.
- Commit: `newtype — type-safe unit wrappers`.

## Goalpost

G7 — Real type checker

EricSpencer00:
Closes #111

## Summary

Adds `newtype Name = BaseType;` for opaque single-field type wrappers — the type-safe counterpart to RES-296 type aliases. Where `type Meters = Float;` makes `Meters` interchangeable with `Float`, `newtype Meters = Float;` makes `Meters` a distinct nominal type. Two newtypes that share a base are NOT interchangeable: `newtype Meters = Float; newtype Volts = Float;` produces two unrelated types, so passing a `Volts` where a `Meters` is expected is a hard compile error.

- Construction via `Name(value)` returns the wrapped value with the new brand.
- Same-brand arithmetic preserves the brand: `Meters + Meters -> Meters`. Cross-brand or newtype-vs-bare-scalar arithmetic is rejected with a diagnostic that names both types.
- Cross-brand assignments and call-argument mismatches surface through the existing `Type::Struct` nominal identity — `Meters` and `Volts` are distinct `Type::Struct(name)` values.
- Runtime erasure — newtypes carry no tag at runtime; the type checker has already proved the program never confuses two newtypes, so the constructor lowers to identity.

## Implementation notes

Follows the feature-isolation pattern from `CLAUDE.md`:
- All feature logic lives in the new `resilient/src/newtypes.rs` (`check`, `is_newtype`, `construct_type`, `infix_result` + 10 unit tests).
- `main.rs`, `lexer_logos.rs`, `typechecker.rs` edits are confined to the `<EXTENSION_*>` blocks (one Token, one keyword, one logos arm, one EXTENSION_PASSES call, plus the AST node).
- Consumer files (`free_vars`, `compiler`, `formatter`, `jit_backend`, `lsp_server`) get one-arm extensions to handle `Node::Newtype` symmetrically with `Node::TypeAlias`.

## Scope notes / follow-ups

- Tuple-style projection `m.0` was considered but deferred — the parser today only accepts identifier field names after `.`, and the ticket's acceptance criteria are satisfied without it. Users who need raw access can model it as a wrapper struct field; once the parser learns numeric tuple indices a tiny follow-up can wire up `.0` projection in this module.
- Comparisons (`==`, `<`, etc.) on newtypes pass through the standard `compatible(&left, &right)` path — same-brand compares as expected, cross-brand reports `Cannot compare Meters and Volts`.

## Test changes

No existing tests modified.

New tests:
- 10 unit tests in `src/newtypes.rs` covering parse, construct, same-brand arithmetic, cross-brand rejection at assignment / call site / arithmetic, duplicate / self-wrap diagnostics, source-position prefix.
- `resilient/examples/newtype_units.rz` + `.expected.txt` golden — exercises Meters / Volts / Seconds, same-brand `Meters + Meters` arithmetic, and a `distance_after(Meters, Meters) -> Meters` function.

### DIFF
--- a/resilient/examples/newtype_units.expected.txt
+++ b/resilient/examples/newtype_units.expected.txt
@@ -1,0 +1,7 @@
+total distance:
+150
+voltage:
+3.3
+elapsed:
+5
+Program executed successfully
--- a/resilient/examples/newtype_units.rz
+++ b/resilient/examples/newtype_units.rz
@@ -1,0 +1,47 @@
+// RES-319: newtype — opaque single-field type wrappers for unit safety.
+//
+// Compare with `type_aliases_units.rz`: that demo uses `type` (transparent
+// alias), so `Meters` and `float` interchange freely. Here we use
+// `newtype` (opaque) so the type checker enforces that `Meters` and
+// `Volts` are distinct types at every use site, even though both wrap
+// `float` underneath.
+//
+// Demonstrated forms:
+//   - top-level `newtype X = T;` declaration
+//   - construction via `Name(value)`
+//   - same-brand arithmetic preserves the brand (`Meters + Meters -> Meters`)
+//   - newtype in a function parameter list
+//
+// Cross-brand operations (`Meters + Volts`, `let m: Meters = v_volts`)
+// are compile-time errors — see the unit-tests in `src/newtypes.rs`
+// for the exact diagnostics.
+
+newtype Meters = float;
+newtype Volts = float;
+newtype Seconds = float;
+
+fn distance_after(Meters start, Meters delta) -> Meters {
+    return start + delta;
+}
+
+fn main() {
+    let m: Meters = Meters(100.0);
+    let v: Volts = Volts(3.3);
+    let t: Seconds = Seconds(5.0);
+
+    // Same-brand arithmetic: `Meters + Meters` returns Meters, and the
+    // result still satisfies the `Meters` parameter binding below.
+    let extra: Meters = Meters(50.0);
+    let total: Meters = distance_after(m, extra);
+
+    // The wrapped values print as their underlying representation;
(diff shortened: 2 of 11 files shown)


EricSpencer00:
Closing for rework — `tests::recursive_function_with_params` (fib(10)) overflows the test thread's stack on Linux CI runners after this change. Passes locally on macOS. Likely cause: the new same-brand arithmetic interception in `InfixExpression` increased typechecker frame size enough to push fib(10) past Linux's default 2MB thread stack. Reopen with one of: (a) run the typechecker on a Builder-thread with bigger stack, (b) replace recursive descent on InfixExpression with an explicit work stack, (c) shrink frame size by boxing newtype fields. Related to #267.

PR #305, opened 2026-04-27
RES-267: cap tree-walker recursion depth

### LINKED ISSUE #267: Tree-walker stack-overflows at ~1000 deep recursion

EricSpencer00:
Closes #267

## Summary

- Adds a tree-walker user-function call-depth guard before another interpreter frame is created.
- Returns a normal runtime error when the guard trips instead of letting deep recursion abort the host process.
- Adds a regression test that evaluates `count(1000)` in a worker thread and asserts the process survives with the expected runtime error.

## Test changes

- Adds `deep_linear_recursion_returns_runtime_error_before_host_overflow`; no existing tests modified.

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml deep_linear_recursion_returns_runtime_error_before_host_overflow`
- [x] `cargo test --manifest-path resilient/Cargo.toml`
- [x] `cargo clippy --manifest-path resilient/Cargo.toml --all-targets -- -D warnings`

### DIFF
--- a/resilient/src/main.rs
+++ b/resilient/src/main.rs
@@ -10142,6 +10142,8 @@ fn sb_struct(buf: String, cap: i64, overflow: bool) -> Value {
 }
 
 // Interpreter for executing Resilient programs
+const MAX_INTERPRETER_CALL_DEPTH: usize = 32;
+
 struct Interpreter {
     env: Environment,
     /// RES-013: static-let bindings. Shared across every sub-interpreter
@@ -10160,6 +10162,9 @@ struct Interpreter {
     /// the runtime check entirely. Empty by default; populated by
     /// `with_proven_fns` after typecheck.
     proven_fns: Rc<HashSet<String>>,
+    /// RES-267: user-function call depth. The tree-walker is recursive
+    /// over function calls; guard it before the host stack overflows.
+    call_depth: usize,
 }
 
 impl Interpreter {
@@ -10174,6 +10179,7 @@ impl Interpreter {
             statics: Rc::new(RefCell::new(HashMap::new())),
             consts: Rc::new(HashMap::new()),
             proven_fns: Rc::new(HashSet::new()),
+            call_depth: 0,
         }
     }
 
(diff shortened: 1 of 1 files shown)


EricSpencer00:
Temporarily closing to unblock #304: both PRs touch resilient/src/main.rs and the overlap guardrail correctly prevents concurrent merges. I will reopen/rebase this after #304 lands.

PR #336, opened 2026-04-29
RES-321: stdlib reference docs — add effect annotations

EricSpencer00:
Closes #282


EricSpencer00:

PR #337, opened 2026-04-29
RES-385a: Linear types — Z3 fallback proof for conditional consumption

### LINKED ISSUE #214: RES-385a: Linear types — Z3 fallback proof for conditional consumption
Follow-up to RES-385 (#203). The MVP pass in `resilient/src/linear.rs` uses a conservative snapshot/restore model across `if` branches: if either branch consumes a linear binding, the value is treated as consumed at the merge point. This rejects some valid programs where *both* branches consume the value but through different control-flow shapes that the static walker can't unify locally.

## Scope

Emit a Z3 obligation in `check_fn_body` when branches disagree on consumption. The verifier must prove that on every live path the linear value is consumed exactly once. Failure to discharge surfaces as a `warning[partial-proof]` (same shape as RES-217).

## Acceptance

- New example `linear_conditional_demo.rz` with `.expected.txt` sidecar exercising the branchy-but-valid shape.
- Unit test under `typechecker::purity_tests` confirming Z3 accepts `if cond { consume(x) } else { consume(x) }`.
- Gated behind `--features z3`; non-Z3 builds keep the current conservative behaviour.

## Out of scope

- Closure capture (see RES-385b).
- Effect-system interaction (see RES-385c).

EricSpencer00:
Closes #214

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,3 +1,8 @@
 {
-  "claims": {}
+  "claims": {
+    "resilient/src/linear.rs": {
+      "branch": "res-385a-linear-conditional-z3",
+      "claimed_at": "2026-04-28T23:55:16Z"
+    }
+  }
 }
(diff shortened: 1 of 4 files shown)


EricSpencer00:

PR #746, opened 2026-05-01
RES-405: generics — VM monomorphization pass (PR 3/4)

EricSpencer00:
## Summary

Implements the VM monomorphization pass for RES-405 (generics).

Closes #368 (partially — PR 3/4, PR 4 adds JIT)

Built on #745 (PR 2/4 — walker plumbing, already merged)

### What ships

- **`resilient/src/monomorph.rs`** — new module with `lower(program: &Node) -> Node`:
  - Phase 1: collect all `fn<T>` declarations
  - Phase 2: collect call sites with literal-typed arguments, infer `T = Int` etc.
  - Phase 3: clone each generic body into `identity$Int`, `identity$String`, etc. and rewrite call sites
  - Original generic declaration is kept as an erasure fallback for variable-typed call sites
- **`lib.rs`** — wire `monomorph::lower` before `compiler::compile` on VM, JIT, and `--disasm` paths
- **`examples/generic_monomorph.rz`** + **`.expected.txt`** — golden sidecar proving `identity(42)`, `identity("hello")`, `identity(true)`, `first(1, "ignored")` all work

### Tests

- 7 new unit tests in `monomorph::tests`:
  - `non_generic_program_passes_through_unchanged`
  - `identity_fn_gets_two_specializations`
  - `mangle_name_single_param` / `mangle_name_two_params`
  - `vm_disasm_shows_specialized_chunks` — disasm output contains `identity$Int` and `identity$String`
  - `monomorphized_program_executes_in_vm`
  - `call_sites_rewritten_to_mangled_names`
- 0 regressions in 2221-test suite

### Design

Per Q3 in `docs/superpowers/specs/2026-04-30-generics-design.md`:
substitution happens post-typecheck in a dedicated lowering pass. Type inference is limited to literals — variable-typed call sites fall through to the generic erasure body, consistent with the tree-walker's behavior.

## Next PRs

- PR 4/4: JIT monomorphization (reuse the VM monomorph pass in `jit_backend.rs`)

PR #747, opened 2026-05-01
RES-405: generics — JIT monomorphization integration (PR 4/4)

### LINKED ISSUE #368: RES-405: Generic monomorphization for fn<T>
## Summary

`fn<T>(x: T) -> T` parses (RES-289) but performs no semantic work — no
substitution, no monomorphization, no specialisation. The typechecker doesn't
validate that `T` is used consistently inside the body; the dynamic interpreter
ignores the type parameter entirely. This blocks the trait system (RES-290).

## Acceptance criteria

- Type parameters `<T, U, ...>` substitute consistently inside the body.
- Calling `id::<Int>(5)` instantiates the body with `T = Int` and produces `Int`.
- Inference: `id(5)` infers `T = Int` without explicit instantiation, by
  unifying argument types.
- Body inconsistency caught: `fn<T> bad(x: T) -> T { return x + 1; }` errors
  because `+ 1` constrains `T = Int` but the signature claims `T` is generic.
- Each instantiation produces a separate compiled body (or, in the interpreter,
  uses the same body but tracks the bound substitution).
- Three golden tests: identity fn (type inferred), swap fn `<T, U>`, mismatched body.
- Commit: `RES-405: generic monomorphization`.

## Notes

- Pairs with RES-290 (traits) and RES-402 (polymorphic containers) — these
  three are the "generics actually work" trio. RES-290 specifically depends
  on this ticket for bound-checking at call sites.
- Lives in a new `resilient/src/monomorph.rs` per the feature-isolation pattern.

## Goalpost

G7 — real type checker.

EricSpencer00:
## Summary

Completes RES-405 generics — JIT backend integration for monomorphization.

Closes #368

Built on #746 (PR 3/4 — VM monomorph pass, already merged).

### What ships

The driver-level `monomorph::lower` call (wired in PR 3 of this series in `lib.rs`) already feeds the monomorphized AST to `jit_backend::run` before JIT compilation. This PR adds integration tests to lock in the behavior:

- `monomorphized_generic_fn_runs_in_jit` — parses `fn identity<T>(T x) -> T { return x; }`, monomorphizes to `identity$Int`, runs through the Cranelift JIT, asserts result = 42
- `monomorphization_preserves_type_param_bounds` — regression guard: verifies the original generic function's `type_param_bounds` field is unchanged after lowering (specialized clones drop their bounds to `[]`, which is correct and tested separately in `monomorph::tests`)

### Tests

2360 tests pass with `--features jit`, 2221 without. No regressions.

### Design notes

The JIT already has two-pass function declaration (Pass 1 declares all `FuncId`s, Pass 2 compiles bodies). After monomorphization, the program contains both the original `identity` generic and the specialized `identity$Int`. Both get declared in Pass 1 and compiled in Pass 2 as regular monomorphic functions. The call site in `main` targets `identity$Int` directly.

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -23,6 +23,10 @@
     "resilient/src/generics.rs": {
       "branch": "res-405-pr3-vm-monomorph",
       "claimed_at": "2026-05-01T00:21:33Z"
+    },
+    "resilient/src/jit_backend.rs": {
+      "branch": "res-405-pr4-jit-monomorph",
+      "claimed_at": "2026-05-01T00:44:05Z"
     }
   }
 }
--- a/resilient/src/jit_backend.rs
+++ b/resilient/src/jit_backend.rs
@@ -4801,4 +4801,53 @@ return add_two(10, 32);
             jit_run_with_symbols(src, &[("add_two", add_two as *const u8, 2)]).expect("jit ok");
         assert_eq!(result, 42, "expected add_two(10,32)=42, got {}", result);
     }
+
+    // RES-405 PR 4: JIT monomorphization — verify the VM monomorph pass
+    // integrates cleanly with the JIT backend.
+
+    #[test]
+    fn monomorphized_generic_fn_runs_in_jit() {
+        // After `monomorph::lower`, `identity(42)` becomes `identity$Int(42)`.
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #812, opened 2026-05-03
RES-779: Associated types design documentation and reference implementation

### LINKED ISSUE #779: RES-775: Effect polymorphism phase 2 — stop defaulting higher-order combinators to @io
Ticket ID: RES-775

Goal
The effect system landed, and RES-193 shipped a best-effort path, but `docs/concurrency.md` still documents a real usability gap: reusable higher-order combinators default to `@io` instead of preserving the callback's effect. This ticket should thread effect variables through higher-order function signatures so a combinator like `map`, `with_retry`, or a callback-taking helper can remain `@pure` when its callback is pure and become `@io` only when the callback is impure.

Files / modules to touch
- resilient/src/typechecker.rs — effect inference / checking for callback parameters and call sites
- resilient/src/generics.rs or related generic-signature plumbing — carry effect variables alongside type params if needed
- resilient/examples/effect_polymorphism.rz — update or replace with additive examples only
- docs/concurrency.md — refresh the "Higher-order effect polymorphism" note when the limitation is gone

Acceptance criteria
- [ ] A higher-order helper that only calls its callback is inferred/checked as preserving the callback's effect instead of always becoming `@io`
- [ ] A pure callback passed through such a helper is accepted in a pure context
- [ ] An impure callback still taints the higher-order helper call site appropriately
- [ ] Diagnostics remain explicit when effect inference cannot resolve a callback's effect
- [ ] New coverage lands without weakening existing effect-system tests

Out of scope
- Full algebraic effect handlers
- Capability-based I/O permissions
- Async / await or preemptive concurrency semantics

Roadmap goal
G9 — Effect system

EricSpencer00:
## Summary
Documents the design and desired syntax for associated types in traits (type members that implementations must define). Includes:
- Comprehensive module-level documentation in traits.rs explaining associated types
- Example trait showing how associated types decouple interface from type choices
- trait_associated_types_design.rz as a reference implementation showing desired syntax
- Implementation roadmap for phase 2 parser/typechecker work

This PR focuses on design documentation without implementing the feature, preparing the way for subsequent PRs that add parser and typechecker support.

## Design Notes
Associated types enable reusable embedded abstractions where type relationships are fixed per implementation. Example:
```rz
trait Transport {
    type Message;
    type Error;
    fn send(self, msg: Self::Message) -> Result<(), Self::Error>;
}
```

This allows each implementation to define its own concrete types while generic code can refer to `T::Message` for type-safe operations.

## Next Steps
1. Parser: support `type Name = ConcreteType;` in impl blocks
2. Typechecker: validate associated type definitions, projection (`T::AssocType`)
3. Generics: carry associated type constraints through monomorphization
4. Examples: demonstrate embedded APIs using associated types

Closes #779

### DIFF
--- a/resilient/examples/trait_associated_types_design.rz
+++ b/resilient/examples/trait_associated_types_design.rz
@@ -1,0 +1,67 @@
+// RES-779: Associated types in traits — design sketch
+// This example shows the desired syntax and behavior for associated types.
+// NOT YET IMPLEMENTED — this is a design reference for RES-779 phase 1-2.
+
+trait Register {
+    // Associated type: the Width type that this register operates on
+    type Width;
+
+    // Method using the associated type
+    fn read(self) -> Self::Width;
+    fn write(self, value: Self::Width);
+}
+
+struct Register8 {
+    addr: int;
+}
+
+struct Register32 {
+    addr: int;
+}
+
+// Implementation for 8-bit register
+impl Register for Register8 {
+    type Width = int;  // 8-bit values
+
+    fn read(self) -> int {
+        return 42;  // Mock read
+    }
+
+    fn write(self, value: int) {
+        // Mock write
+    }
+}
+
+// Implementation for 32-bit register
+impl Register for Register32 {
+    type Width = int;  // 32-bit values (same type, but semantically different width)
+
+    fn read(self) -> int {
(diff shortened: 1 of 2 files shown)


EricSpencer00:

PR #1090, opened 2026-05-09
Multi-bug pass: fix six ship-blocking issues (#1079-1083, #1087)

EricSpencer00:
## Summary
Fixes six independently-filed ship-blocking issues in one pass. Each
fix is small and orthogonal — they're bundled here because they all
land in the lexer / interpreter / examples and would otherwise produce
six near-identical PRs touching the same files.

| Issue | Fix |
|---|---|
| [#1079](https://github.com/EricSpencer00/Resilient/issues/1079) | `asin` precision — replace `sin(asin(0.5))` round-trip with `abs(round_trip - 0.5) < 1e-9` |
| [#1080](https://github.com/EricSpencer00/Resilient/issues/1080) | `asinh` precision — same pattern |
| [#1081](https://github.com/EricSpencer00/Resilient/issues/1081) | `atanh` precision — same pattern |
| [#1082](https://github.com/EricSpencer00/Resilient/issues/1082) | Octal `0o…` literals — add regex + callback in both lexers |
| [#1083](https://github.com/EricSpencer00/Resilient/issues/1083) | Nested block comments — track depth; return `Eof` on unterminated |
| [#1087](https://github.com/EricSpencer00/Resilient/issues/1087) | Recursion limit — 32 → 256 default, `RZ_MAX_CALL_DEPTH` env override |

## Test plan
- [x] `cargo build --manifest-path resilient/Cargo.toml --release`
- [x] `cargo test --manifest-path resilient/Cargo.toml --release --lib` — **2687 passed, 0 failed**
- [x] `cargo test --manifest-path resilient/Cargo.toml --release --test examples_smoke` — passed
- [x] `cargo test --manifest-path resilient/Cargo.toml --release --test examples_golden` — passed (validates all `.expected.txt` against actual `.rz` output)
- [x] `cargo fmt --manifest-path resilient/Cargo.toml --check` — clean
- [x] `cargo clippy --manifest-path resilient/Cargo.toml --release --all-targets -- -D warnings` — clean
- [x] Manual verification of all six fixes via the CLI binary

## Test changes
- Added two new unit tests (`octal_literals`, `nested_block_comments`) — additive, no edits to existing tests.
- Updated `deep_linear_recursion_returns_runtime_error_before_host_overflow` to use `count(1024)` so it still triggers the depth-guard now that the default is 256 instead of 32. The test's *safety property* (runaway recursion errors rather than crashes) is unchanged.
[...]

PR #1114, opened 2026-05-11
RES-1104..1113: ten-bug pass — typechecker scope + equality + block scope + reachability

### LINKED ISSUE #1104: RES-1104: typechecker — for-in loop variable false 'Undefined variable' error
## Bug

The typechecker reports false-positive `Undefined variable` for the binding introduced by a `for-in` loop. The runtime works correctly; only the soft-mode typechecker complains, which pollutes diagnostics and prevents `--typecheck` strict mode from passing.

## Repro

\`\`\`rust
fn main() {
    let xs = [0, 1, 2, 3, 4];
    let total = 0;
    for x in xs {
        total = total + x;   // typechecker: \"Undefined variable 'x'\"
    }
    println(total);
}
main();
\`\`\`

Output:
\`\`\`
Type error: Undefined variable 'x' at 5:26
10
Program executed successfully
\`\`\`

Strict typecheck (\`--typecheck\`) fails outright.

## Root cause

\`Node::ForInStatement\` in \`resilient/src/typechecker.rs\` checks the iterable then recurses into the body without binding the loop variable name into the typechecker environment. The desugared tuple-for case (\`\$for_tup\$N\`) has the same gap.

## Fix sketch

Bind \`name\` to \`Type::Any\` (or an element type derived from the iterable) in a fresh enclosed env before checking the body; restore the outer env afterwards.

## Acceptance

- \`for x in [1,2,3] { println(x); }\` typechecks clean
- \`for i in 0..5 { println(i); }\` typechecks clean
- \`for (k, v) in pairs { ... }\` (tuple binding) still typechecks
- New regression tests in \`typechecker.rs\` covering the three forms

### LINKED ISSUE #1105: RES-1105: typechecker — recursive function calls fail 'Undefined variable'
## Bug

The typechecker reports \`Undefined variable\` when a function calls itself. The runtime handles recursion correctly; only the soft-mode typechecker errors, blocking \`--typecheck\` strict mode for any recursive function.

## Repro

\`\`\`rust
fn fact(int n) -> int {
    if n <= 1 { return 1; }
    return n * fact(n - 1);   // typechecker: \"Undefined variable 'fact'\"
}

fn main() {
    println(fact(5));
}
main();
\`\`\`

Output:
\`\`\`
Type error: Undefined variable 'fact' at 3:20
120
Program executed successfully
\`\`\`

## Root cause

In \`typechecker.rs::check_program_with_source\`, the pre-pass populates \`contract_table\` and \`fn_decl_spans\` for forward references but does NOT register the function name into \`self.env\`. The function name is added to \`env\` only AFTER its body is checked. Identifier lookup goes through \`env.get\`, so the body sees no \`fact\`.

## Fix sketch

In the pre-pass, also insert each function into \`self.env\` with a \`Type::Function { params, return_type }\` derived from the declared parameter types and return annotation. After the body check, refresh the entry with the actual (possibly inferred) return type.

## Acceptance

- \`fact\`-style direct self-recursion typechecks clean
- The Z3 contract pre-pass and \`fn_decl_spans\` book-keeping remain functional
- New regression tests covering direct recursion in \`typechecker.rs\`

### LINKED ISSUE #1106: RES-1106: typechecker — mutual / forward function calls fail 'Undefined variable'
## Bug

Forward references between functions fail the typechecker, contradicting SYNTAX.md's promise that \"Functions can be defined in any order — forward references work\". The runtime resolves the forward reference correctly; only the typechecker errors.

## Repro

\`\`\`rust
fn even(int n) -> bool {
    if n == 0 { return true; }
    return odd(n - 1);   // typechecker: \"Undefined variable 'odd'\"
}
fn odd(int n) -> bool {
    if n == 0 { return false; }
    return even(n - 1);
}
fn main() {
    println(even(4));
}
main();
\`\`\`

Output:
\`\`\`
Type error: Undefined variable 'odd' at 3:15 — did you mean \`ord\`, \`gcd\`?
true
Program executed successfully
\`\`\`

## Root cause

Same as RES-1105: the typechecker's pre-pass populates \`contract_table\` but never inserts function names into \`self.env\`. So the first function's body cannot see later-declared functions, even though the contract pre-pass already promised they exist.

## Fix sketch

Same fix as RES-1105 — extend the pre-pass to also register function names in \`self.env\` with their declared signatures. Both issues share one fix.

## Acceptance

- Mutual recursion (\`even\`/\`odd\`) typechecks clean
- A function calling a sibling declared textually below typechecks clean
- Regression tests covering both

### LINKED ISSUE #1107: RES-1107: runtime — short-circuit '&&' and '||' evaluate both sides
## Bug

The logical operators \`&&\` and \`||\` evaluate **both** operands at runtime, in contradiction of every standard short-circuit language (C, Rust, Python, JS, Java, Go). This is a correctness hazard for any guard expression like \`x != 0 && y / x > 0\` — the right side runs even when the left is false.

## Repro

\`\`\`rust
fn side(int x) -> bool { println(\"called\"); return x > 0; }
fn main() {
    let r = false && side(1);
    println(r);
    let r2 = true || side(2);
    println(r2);
}
main();
\`\`\`

Output (both \"called\" lines should NOT appear):
\`\`\`
called
false
called
true
\`\`\`

Expected:
\`\`\`
false
true
\`\`\`

## Root cause

\`Node::InfixExpression\` in the tree-walker eagerly evaluates \`left\` and \`right\` before dispatching on the operator, so \`&&\` / \`||\` are pure value-based after both sides have run.

## Fix sketch

In the eval arm for \`Node::InfixExpression\`, intercept \`&&\` / \`||\` BEFORE evaluating \`right\`:
- \`&&\`: eval left; if \`false\`, return \`false\`; else eval right
- \`||\`: eval left; if \`true\`, return \`true\`; else eval right
- Non-bool left falls through to the standard error path

## Acceptance

- \`false && side(1)\` does NOT call \`side\`
- \`true || side(2)\` does NOT call \`side\`
- Non-bool operands still error with the existing diagnostic
- Regression test in \`lib.rs#tests\` and a golden example

### LINKED ISSUE #1108: RES-1108: runtime — Array '==' / '!=' panics with 'Type mismatch'
## Bug

Comparing two arrays with \`==\` or \`!=\` at runtime returns a \"Type mismatch\" error even though the typechecker accepts the expression. Standard equality between structurally identical arrays should return \`true\`.

## Repro

\`\`\`rust
fn main() {
    let a = [1, 2, 3];
    let b = [1, 2, 3];
    println(a == b);   // Runtime error: Type mismatch
}
main();
\`\`\`

Output:
\`\`\`
Runtime error: Type mismatch: [1, 2, 3] == [1, 2, 3]
\`\`\`

## Root cause

\`Interpreter::eval_infix_expression\` dispatches on type pairs and falls through to a \"Type mismatch\" error when neither side is one of \`Int\`/\`Float\`/\`String\`/\`Bool\`. There's no Array == Array case.

## Fix sketch

Before the fallthrough \`Type mismatch\`, add Array == Array (length + element-wise compare) and \`!=\` as the inverse. Use \`Value\`'s \`PartialEq\` for element comparison.

## Acceptance

- \`[1,2,3] == [1,2,3]\` returns \`true\`
- \`[1,2] == [1,2,3]\` returns \`false\` (length mismatch)
- \`[\"a\"] == [\"a\"]\` returns \`true\`
- \`[[1],[2]] == [[1],[2]]\` returns \`true\` (nested)
- \`!=\` is the inverse
- Regression test + golden example

### LINKED ISSUE #1109: RES-1109: runtime — Struct '==' / '!=' panics with 'Type mismatch'
## Bug

Comparing two struct values with \`==\` / \`!=\` at runtime errors with \"Type mismatch\". Structural equality between identical struct values should return \`true\`.

## Repro

\`\`\`rust
struct Point { int x, int y, }
fn main() {
    let a = new Point { x: 1, y: 2 };
    let b = new Point { x: 1, y: 2 };
    println(a == b);   // Runtime error: Type mismatch
}
main();
\`\`\`

## Root cause

Same code path as RES-1108. \`eval_infix_expression\` lacks a Struct == Struct branch.

## Fix sketch

Add Struct == Struct comparison: same struct name + same field set + element-wise field equality. Different struct names compare as \`false\` (and \`!=\` returns \`true\`).

## Acceptance

- Identical \`Point\` values compare equal
- \`Point{x:1,y:2}\` ≠ \`Point{x:1,y:3}\`
- Different struct types compare \`false\` (no panic)
- Regression test + golden example

### LINKED ISSUE #1110: RES-1110: runtime — Tuple '==' / '!=' panics with 'Type mismatch'
## Bug

Comparing two tuple values with \`==\` / \`!=\` at runtime errors with \"Type mismatch\". Resilient already supports tuple destructuring everywhere (\`let (a, b) = t;\`), tuple match patterns, and tuple binding in \`for\`. Equality should round out the set.

## Repro

\`\`\`rust
fn main() {
    let t1 = (1, 2);
    let t2 = (1, 2);
    println(t1 == t2);   // Runtime error: Type mismatch
}
main();
\`\`\`

## Root cause

Same code path as RES-1108. \`eval_infix_expression\` has no Tuple == Tuple branch.

## Fix sketch

Add Tuple == Tuple comparison: same arity + element-wise equality. Different arities compare \`false\`.

## Acceptance

- \`(1,2) == (1,2)\` returns \`true\`
- \`(1,2) == (1,3)\` returns \`false\`
- \`(1,2) == (1,2,3)\` returns \`false\` (arity mismatch, no panic)
- Regression test + golden example

### LINKED ISSUE #1111: RES-1111: runtime — block-scoped 'let' leaks into the enclosing scope
## Bug

\`let\` declarations inside an \`if\` / \`while\` / \`for\` body persist in the **enclosing** scope after the block exits. This violates standard lexical-scope semantics (Rust, C, JavaScript \`let\`, etc.) and is a serious safety hazard for a safety-critical language: an inner-scope rebinding silently mutates outer bindings.

## Repro 1 — outer 'x' clobbered

\`\`\`rust
fn main() {
    let x = 1;
    if true {
        let x = 99;
        println(x);  // 99 (correct — inner scope)
    }
    println(x);      // expected 1, got 99
}
main();
\`\`\`

## Repro 2 — inner 'm' visible after the loop

\`\`\`rust
fn main() {
    let n = 1;
    while n < 3 {
        let m = 100;
        n = n + 1;
    }
    println(m);      // 100 — should be a scope error
}
main();
\`\`\`

## Root cause

\`Interpreter::eval_block_statement\` (\`lib.rs\`) iterates statements directly against the current \`self.env\` without pushing/popping a scope frame. So every \`let\` inside a block is bound at the function-frame level.

The typechecker DOES create a new enclosed env per block (\`Node::Block\`), which is why Repro 2's outer \`println(m)\` produces a soft-mode type error — but the runtime still prints \`100\`.

## Fix sketch

In \`eval_block_statement\`, push a new enclosed env on entry and restore the outer env on exit. Mirror the typechecker's existing pattern. Existing semantics for \`return\` / \`break\` / \`continue\` propagation must be preserved.

Care: many existing examples rely on \`live\` blocks and similar constructs that may not be \`Node::Block\`. Inspect carefully — only \`Node::Block\` needs the new scope frame; live-block snapshots are a separate mechanism.

## Acceptance

- Repro 1: outer \`println(x)\` prints \`1\`
- Repro 2: \`println(m)\` raises an \"Identifier not found: m\" runtime error
- All existing examples still pass golden checks
- New regression tests for both forms

### LINKED ISSUE #1112: RES-1112: typechecker — function declared '-> T' falling off the end is not flagged
## Bug

A function declared to return a non-\`void\` type is allowed to silently fall off the end. The typechecker accepts it, the runtime returns \`Value::Void\`, and downstream code reading the result observes a type that contradicts the declared signature.

## Repro

\`\`\`rust
fn maybe(int n) -> int {
    if n > 0 { return 1; }
    // no return on the else path
}
fn main() {
    println(maybe(5));     // 1
    println(maybe(-1));    // void  ← should never happen for -> int
}
main();
\`\`\`

Output (under both default and \`--typecheck\`):
\`\`\`
Type check passed
1
void
\`\`\`

## Root cause

The typechecker only checks the **last** statement's type against the declared return type. It does not verify that every control-flow path ends in a \`return\` (or a diverging construct).

## Fix sketch

Add a reachability/completeness analysis: a fn body is \"complete\" for type \`T\` if every leaf of its control-flow graph is one of:
- \`return EXPR\` with \`EXPR : T\`
- a call to a diverging function (e.g. \`panic\`, \`unreachable\`)
- an infinite loop without \`break\`

Otherwise emit \`fn NAME: missing return on path — declared \`T\`, but body falls off the end\`.

Scope guard: ship the **conservative** version first — only flag the common case (\`if\` without \`else\`, \`match\` with non-exhaustive arms covered by RES-117). Subtle cases (loops, try/catch) can land later.

## Acceptance

- The repro is rejected by \`--typecheck\`
- A function whose every branch \`return\`s typechecks clean
- \`void\`-returning functions are unaffected
- Regression tests covering: if-without-else, if/elif without else, all-branches-return

### LINKED ISSUE #1113: RES-1113: typechecker — code after 'return' is silently accepted (unreachable)
## Bug

Statements after an unconditional \`return\` are not flagged. This is a common author error (left-over debug code, accidentally orphaned cleanup) and every static-analyzed language warns or errors on it.

## Repro

\`\`\`rust
fn classify(int n) -> string {
    if n > 0 { return \"pos\"; }
    return \"nonpos\";
    return \"dead\";          // unreachable, silently accepted
}
fn main() {
    println(classify(5));
}
main();
\`\`\`

Strict typecheck still passes; the second \`return\` is silently dropped.

\`\`\`rust
fn main() {
    return;
    println(\"unreachable\");   // also silently accepted
}
main();
\`\`\`

## Root cause

The typechecker walks every statement in a block and checks each in isolation. It does not maintain a \"reachable\" flag that flips to \`false\` after a terminating statement.

## Fix sketch

In \`check_node\` for \`Node::Block\` (or in a new lint pass), track an \`is_reachable\` flag:
- A \`return\`, \`break\`, or \`continue\` terminates the rest of the block
- The first unreachable statement is flagged with \`error: unreachable code after \`return\` (or \`break\`/\`continue\`)\`

This can be a hard error (consistent with safety-critical posture) or a warning behind \`--deny-warnings\`. Default to hard error since dead code is unambiguous.

## Acceptance

- The repro rejects with \`unreachable code after return\`
- \`break\` / \`continue\` in loop bodies trigger the same diagnostic
- Code reachable via at least one path (e.g. after a non-terminating \`if\`) is NOT flagged
- Regression tests for: post-return, post-break, post-continue, and the negative case

EricSpencer00:
Closes #1104. Closes #1105. Closes #1106. Closes #1107. Closes #1108.
Closes #1109. Closes #1110. Closes #1111. Closes #1112. Closes #1113.

Ten high-quality bugs in standard language features, all caught by
deliberately probing for-in loops, recursion, short-circuit operators,
compound equality, lexical scoping, and control-flow completeness.
Bundled into one PR because the touch points cluster in `lib.rs` +
`typechecker.rs` and the test infrastructure is identical (matches
the recent multi-bug pattern in #1090 / #1091 / #1103).

## Typechecker scope (RES-1104, RES-1105, RES-1106)

| Before | After |
|--|--|
| `for x in xs { ... }` → `Undefined variable 'x'` | typechecks clean |
| `fn fact(n) -> int { ... fact(n-1) ... }` → `Undefined variable 'fact'` | typechecks clean |
| `fn even() { odd() }; fn odd() { even() }` → `Undefined variable 'odd'` | typechecks clean |

Root cause: the for-in arm never pushed the loop binding into the env, and the function pre-pass only filled `contract_table`, not `self.env`. Both fixed by enclosing the body in a fresh env that carries the binding.

## Runtime correctness (RES-1107..RES-1110)

| Operator | Before | After |
|--|--|--|
| `false && side(1)` | `side` runs, returns `false` | `side` is **not** called, returns `false` |
| `[1,2,3] == [1,2,3]` | runtime panic "Type mismatch" | `true` |
| `Point{1,2} == Point{1,2}` | runtime panic "Type mismatch" | `true` |
| `(1,2) == (1,2)` | runtime panic "Type mismatch" | `true` |

Compound equality is recursive (nested arrays / tuples / structs compare element-wise) and total once you've crossed the compound boundary — mismatched shapes return `false` rather than panic.

## Runtime scope (RES-1111)

```rust
fn main() {
    let x = 1;
    if true {
        let x = 99;
    }
    println(x);   // before: 99 (!),  after: 1
}
```

`eval_block_statement` now pushes a fresh enclosed env per block, mirroring the typechecker. Reassignment (`x = 5`) still walks the outer chain so updating an outer binding from inside a block keeps working.

## Static analysis (RES-1112, RES-1113)

```rust
// RES-1112: missing return
fn maybe(int n) -> int {
    if n > 0 { return 1; }
}  // ERROR: missing return on at least one path
```

```rust
// RES-1113: unreachable code
fn classify(int n) -> string {
    return "a";
    return "b";  // ERROR: unreachable code
}
```

`node_terminates` recurses through `Block`, `If` (with else), `Match` (when every arm returns), `LiveBlock`, and `TryCatch`. The implicit-return form (`fn id(int x) -> int { x }`) is recognised via `body_yields_value` so it doesn't false-positive.

## Tests

- 14 new typechecker unit tests in `scope_and_reachability_tests` cover the positive and negative cases for RES-1104..1106 and RES-1112..1113. Includes the implicit-return form and the `live { return X }` shape used by real safety-critical examples (verified by typechecking all bundled examples — zero new failures after the helper recognised `LiveBlock` / `TryCatch` as terminating).
- 5 new golden examples (`res1107_short_circuit.rz` through `res1111_block_scope.rz`) exercise the runtime fixes end-to-end via the existing `examples_golden` harness.
- All 2703 existing lib tests pass; full integration suite passes; clippy + fmt clean; cortex-m4 cross-compile passes.

## Local CI parity

- `cargo build --locked` ✅
- `cargo test --locked` ✅ (2703 lib + integration suites all green)
- `cargo clippy --all-targets -- -D warnings` ✅
- `cargo fmt --check` ✅
- `cargo build --target thumbv7em-none-eabihf` ✅
- Z3 path not exercised locally (no z3-dev installed) — CI gate covers it.

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,8 +1,12 @@
 {
   "claims": {
     "resilient/src/lib.rs": {
-      "branch": "res-934-as-cast",
-      "claimed_at": "2026-05-06T06:31:27Z"
+      "branch": "claude/wizardly-feynman-0204f0",
+      "claimed_at": "2026-05-11T01:39:51Z"
+    },
+    "resilient/src/typechecker.rs": {
+      "branch": "claude/wizardly-feynman-0204f0",
+      "claimed_at": "2026-05-11T01:39:51Z"
     }
   }
 }
--- a/resilient/examples/res1107_short_circuit.expected.txt
+++ b/resilient/examples/res1107_short_circuit.expected.txt
@@ -1,0 +1,7 @@
+false
+true
+called
+true
+called
+true
+Program executed successfully
--- a/resilient/examples/res1107_short_circuit.rz
+++ b/resilient/examples/res1107_short_circuit.rz
@@ -1,0 +1,28 @@
+// RES-1107: `&&` and `||` short-circuit. The right-hand side is only
+// evaluated when its result is required to determine the operator's
+// outcome. This is essential for guard idioms like
+// `x != 0 && y / x > 0`.
+
+fn side(int x) -> bool {
+    println("called");
+    return x > 0;
+}
+
+fn main() {
+    // `false && _` returns false WITHOUT touching the right side.
+    let a = false && side(1);
+    println(a);
+
+    // `true || _` returns true WITHOUT touching the right side.
+    let b = true || side(2);
+    println(b);
+
+    // Both sides DO run when the left is the not-yet-determined value.
+    let c = true && side(3);
+    println(c);
+
+    let d = false || side(4);
+    println(d);
+}
+
+main();
--- a/resilient/examples/res1108_array_equality.expected.txt
+++ b/resilient/examples/res1108_array_equality.expected.txt
@@ -1,0 +1,9 @@
+true
+false
+false
+true
+true
+true
+false
+true
+Program executed successfully
--- a/resilient/examples/res1108_array_equality.rz
+++ b/resilient/examples/res1108_array_equality.rz
@@ -1,0 +1,29 @@
+// RES-1108: structural `==` / `!=` for arrays. Same length + element-
+// wise equality. Different lengths compare unequal rather than panic.
+
+fn main() {
+    let a = [1, 2, 3];
+    let b = [1, 2, 3];
+    let c = [1, 2, 4];
+    let d = [1, 2];
+
+    println(a == b);    // true  — structurally identical
+    println(a == c);    // false — element diff
+    println(a == d);    // false — length diff
+    println(a != c);    // true  — inverse of ==
+    println([] == []);  // true  — empty arrays
+
+    // Nested arrays compare recursively.
+    let nested1 = [[1, 2], [3, 4]];
+    let nested2 = [[1, 2], [3, 4]];
+    let nested3 = [[1, 2], [3, 5]];
+    println(nested1 == nested2);   // true
+    println(nested1 == nested3);   // false
+
+    // Mixed-element arrays still compare element-wise.
+    let mixed1 = ["a", "b"];
+    let mixed2 = ["a", "b"];
+    println(mixed1 == mixed2);     // true
+}
+
+main();
--- a/resilient/examples/res1109_struct_equality.expected.txt
+++ b/resilient/examples/res1109_struct_equality.expected.txt
@@ -1,0 +1,5 @@
+true
+false
+false
+true
+Program executed successfully
--- a/resilient/examples/res1109_struct_equality.rz
+++ b/resilient/examples/res1109_struct_equality.rz
@@ -1,0 +1,20 @@
+// RES-1109: structural `==` / `!=` for struct values. Same struct
+// name + same field set + element-wise field equality. Different
+// struct names compare unequal rather than panic.
+
+struct Point { int x, int y, }
+struct Other { int x, int y, }
+
+fn main() {
+    let a = new Point { x: 1, y: 2 };
+    let b = new Point { x: 1, y: 2 };
+    let c = new Point { x: 1, y: 3 };
+    let d = new Other { x: 1, y: 2 };
+
+    println(a == b);    // true  — identical
+    println(a == c);    // false — field diff
+    println(a == d);    // false — different struct names
+    println(a != c);    // true  — inverse of ==
+}
+
+main();
--- a/resilient/examples/res1110_tuple_equality.expected.txt
+++ b/resilient/examples/res1110_tuple_equality.expected.txt
@@ -1,0 +1,8 @@
+true
+false
+true
+true
+false
+true
+false
+Program executed successfully
--- a/resilient/examples/res1110_tuple_equality.rz
+++ b/resilient/examples/res1110_tuple_equality.rz
@@ -1,0 +1,23 @@
+// RES-1110: structural `==` / `!=` for tuples. Same arity +
+// element-wise equality. Different arities compare unequal rather
+// than panic.
+
+fn main() {
+    let a = (1, 2);
+    let b = (1, 2);
+    let c = (1, 3);
+
+    println(a == b);            // true
+    println(a == c);            // false
+    println(a != c);            // true
+
+    // 3-tuples
+    println((1, 2, 3) == (1, 2, 3));    // true
+    println((1, 2, 3) == (1, 2, 4));    // false
+
+    // Heterogeneous element types
+    println((1, "x") == (1, "x"));      // true
+    println((1, "x") == (1, "y"));      // false
+}
+
+main();
--- a/resilient/examples/res1111_block_scope.expected.txt
+++ b/resilient/examples/res1111_block_scope.expected.txt
@@ -1,0 +1,5 @@
+99
+1
+10
+6
+Program executed successfully
--- a/resilient/examples/res1111_block_scope.rz
+++ b/resilient/examples/res1111_block_scope.rz
@@ -1,0 +1,32 @@
+// RES-1111: `let` declarations inside an `if` / `while` / `for` body
+// are confined to that body. Without this rule a stray inner `let`
+// silently clobbered outer bindings — a serious hazard for safety-
+// critical code.
+
+fn main() {
+    // The outer `x` survives an inner-scope rebinding.
+    let x = 1;
+    if true {
+        let x = 99;
+        println(x);    // 99  (inner scope)
+    }
+    println(x);        // 1   (outer scope intact)
+
+    // Reassignment from inside a block still updates the outer
+    // binding, since `=` walks the env chain.
+    let counter = 0;
+    if true {
+        counter = counter + 10;
+    }
+    println(counter);  // 10
+
+    // Loop body bindings are confined to the body.
+    let total = 0;
+    for i in 0..3 {
+        let doubled = i * 2;
+        total = total + doubled;
+    }
+    println(total);    // 0+2+4 = 6
+}
+
+main();
--- a/resilient/src/lib.rs
+++ b/resilient/src/lib.rs
@@ -20231,6 +20231,87 @@ struct Interpreter {
     inject_checked_failures: bool,
 }
 
+/// RES-1108 + RES-1109 + RES-1110: structural equality for compound
+/// values (Array, Tuple, Struct). Returns `Some(true)` / `Some(false)`
+/// when both operands are the same compound shape. Returns `None` when
+/// the operands are not both compounds — in that case the caller falls
+/// through to the existing primitive / mismatch dispatch so cross-type
+/// comparisons (`[1] == "x"`) still error.
+///
+/// Recursion uses `values_strict_eq` which treats any type mismatch
+/// inside the recursion as `false` rather than an error, so element-wise
+/// comparisons inside arrays / tuples are total.
+fn compound_values_equal(left: &Value, right: &Value) -> Option<bool> {
+    match (left, right) {
+        (Value::Array(l), Value::Array(r)) => Some(slices_strict_eq(l, r)),
+        (Value::Tuple(l), Value::Tuple(r)) => Some(slices_strict_eq(l, r)),
+        (
+            Value::Struct {
+                name: ln,
+                fields: lf,
+            },
+            Value::Struct {
+                name: rn,
+                fields: rf,
+            },
+        ) => {
+            if ln != rn {
+                return Some(false);
+            }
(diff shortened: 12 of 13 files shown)


EricSpencer00:

PR #1187, opened 2026-05-12
RES-1186: per-OS golden sidecars to fix macOS expm1(1.0) flake

### LINKED ISSUE #1186: RES-1186: precision_math golden flakes on macOS — f64::exp_m1(1.0) libc divergence
## Problem

`resilient/tests/examples_golden.rs::golden_outputs_match` is **deterministically red on macOS** for `examples/precision_math.rz` because `f64::exp_m1(1.0)`:

- on Linux glibc (CI): `1.718281828459045`  ← golden was generated here
- on macOS libsystem_m (local): `1.7182818284590453`

Same input bits, different shortest-round-trippable decimal — the underlying libc `expm1` implementations differ in their last ULP for `x = 1.0`. Reproduced on stock `origin/main` with no other changes, so this is **not caused by RES-1184**.

Net effect: every Mac developer's `agent-scripts/verify-scope.sh` is permanently red, which blocks `ready-or-bail.sh` from marking unrelated PRs ready. The diff being shipped doesn't touch `precision_math.rs` or the math builtins.

## Proposal options (lowest-cost first)

1. **Replace `expm1(1.0)` in the example with a value that's bit-exact across libcs** (e.g. `expm1(0.0)` is already in the file; add a comment explaining why we don't test `expm1(1.0)` for output). Then add a unit test inside `precision_math.rs::tests` that asserts the macOS-vs-Linux invariant we DO care about (relative error vs naive `exp(x) - 1`), which can use `assert!(close(..))` instead of byte-equal stdout.
2. **Per-platform expected file**: `precision_math.expected.linux.txt` / `precision_math.expected.macos.txt`, with the harness preferring the platform-specific sidecar when present and falling back to `.expected.txt`. ~15 LoC in `tests/examples_golden.rs`.
3. **Tolerance-based comparator** for float lines in the golden harness (parse-and-compare floats with epsilon, byte-compare everything else). Bigger surface but useful for every future precision-sensitive demo.

(1) is the smallest fix and ships in one PR. (2) is the most honest fix and keeps the language-level precision demo intact across platforms. (3) is the long-term right answer but is its own scoping discussion.

## Acceptance

- `cargo test --manifest-path resilient/Cargo.toml --test examples_golden golden_outputs_match` passes on both Linux and macOS hosts.
- Unit tests in `resilient/src/precision_math.rs` continue to assert the precision invariant (whichever option is taken).
- No regression in any other golden.

EricSpencer00:
## Summary

`examples_golden::golden_outputs_match` is deterministically red on macOS for `examples/precision_math.rz` because `f64::exp_m1(1.0)`:

- on Linux glibc (CI): `1.718281828459045` — what the golden expected
- on macOS libsystem_m: `1.7182818284590453` — what mac devs see

Same input bits, different shortest round-trippable decimal — the underlying libc `expm1` implementations diverge in their last ULP. Reproduces on stock `origin/main` with no other changes, so this is a portability bug introduced by RES-1168, not by any recent profile/perf work. Every Mac developer's `agent-scripts/verify-scope.sh` has been red since RES-1168 merged, silently blocking `ready-or-bail.sh` for every unrelated PR (including RES-1184).

## Approach

Smallest reasonable fix: `expected_path()` in `resilient/tests/examples_golden.rs` now looks for a sibling `<stem>.expected.<os>.txt` (`os` = `std::env::consts::OS`, e.g. `macos`, `linux`, `windows`) before falling back to the existing `<stem>.expected.txt`. The Linux value stays canonical in the default sidecar; macOS gets its own `precision_math.expected.macos.txt` carrying the libsystem_m output. Any future precision-sensitive demo that drifts across platforms can add a per-OS override the same way without touching the harness.

15-line harness change + one 12-line sidecar. No behavioural change on Linux (the default `.expected.txt` is still the only file that matches there). No new dependencies, no API surface, no changes to any other test.

## Scope

- `resilient/tests/examples_golden.rs` — `expected_path()` gains the platform-specific lookup.
- `resilient/examples/precision_math.expected.macos.txt` — new sidecar for the macOS libsystem_m float-rounding case.

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml --test examples_golden golden_outputs_match` — passes on macOS (verified locally).
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, full `cargo test` — full guardrail green locally.
- [ ] Linux CI confirms the default `.expected.txt` path stays load-bearing.

Closes #1186


EricSpencer00:

PR #1336, opened 2026-05-12
RES-1334: skip source-file SHA-256 + cache::check when --no-cache is set

### LINKED ISSUE #1334: RES-1334: skip source-file SHA-256 + cache::check when --no-cache is set
\`lib.rs:25736\` unconditionally computes a SHA-256 of the source file contents to feed into \`cache::check\` (the RES-355 incremental-bytecode-cache lookup). The cache::check call itself is then short-circuited on \`!no_cache\` at line 25746 — so when \`--no-cache\` is passed, the SHA is computed and never used.

\`\`\`rust
let source_hash = {
    use sha2::{Digest, Sha256};
    let mut hasher = Sha256::new();
    hasher.update(contents.as_bytes());
    format!("{:x}", hasher.finalize())
};
let cache_dir = cache::cache_dir_for(filename);
// no_cache=true → cache::check skipped, _cache_hit always false
let _cache_hit = !no_cache && cache::check(&cache_dir, &source_hash) == cache::CacheResult::Hit;
\`\`\`

\`source_hash\` is consumed only by \`cache::check\` (grep confirms no other use). When \`--no-cache\` flips the short-circuit, the SHA is dead work — typical full-source hashes are microseconds, large files (10K+ lines) push into milliseconds, and every \`cargo run --no-cache\` / test invocation that sets the flag pays the cost.

Gate the SHA computation on \`!no_cache\` so it runs only when the cache lookup will actually consume it. Behaviour is unchanged for the default code path; \`--no-cache\` invocations skip the hash entirely.

Out of scope: the broader question of whether the \`_cache_hit\` discard pattern (the cache::check whose result is currently never read; comment claims \"future phase can branch on it\") deserves removal. Keep that intact — only the SHA prerequisite is touched here.

EricSpencer00:
Closes #1334

## Summary

`lib.rs::run_program_with_args` unconditionally computed a SHA-256 of the entire source file and called `cache::check` / `cache::write_entry` even when `--no-cache` was passed. Both consumers short-circuit on the `no_cache` flag, so the hash was dead work — full-source SHA-256 is microseconds for small files and pushes into milliseconds for 10k+-line inputs.

Bind `source_hash` and `cache_dir` as `Option`s populated only when `!no_cache`. The three downstream `cache::write_entry` sites (JIT, VM, tree-walker paths) match on the pair before writing — keeping the existing skip-on-no-cache behaviour with no extra branches.

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml` (full suite green, including the `cache::` smoke tests that exercise both the `--cache` and `--no-cache` paths)
- [x] `cargo clippy --manifest-path resilient/Cargo.toml --all-targets -- -D warnings` clean

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -189,8 +189,8 @@
       "claimed_at": "2026-05-12T05:37:55Z"
     },
     "resilient/src/lib.rs": {
-      "branch": "res-1332-skip-lints-when-not-safety-critical",
-      "claimed_at": "2026-05-12T06:22:22Z"
+      "branch": "res-1334-skip-sha-when-no-cache",
+      "claimed_at": "2026-05-12T06:30:36Z"
     }
   }
 }
--- a/resilient/src/lib.rs
+++ b/resilient/src/lib.rs
@@ -25733,17 +25733,33 @@ fn execute_file(
     // RES-1202: hash inline (was `cert_sign::sha256_hex`) so the
     // incremental cache stays independent of the z3-feature-gated
     // `cert_sign` module.
-    let source_hash = {
-        use sha2::{Digest, Sha256};
-        let mut hasher = Sha256::new();
-        hasher.update(contents.as_bytes());
-        format!("{:x}", hasher.finalize())
+    //
+    // RES-1334: gate the SHA-256 + cache lookup on `!no_cache`. The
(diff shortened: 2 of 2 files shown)


EricSpencer00:
Closed in favor of #1335, which shipped the same RES-1334 fix first.

PR #1359, opened 2026-05-12
RES-1357: gate self.certificates push on opt-in capture_certificates flag

### LINKED ISSUE #1357: RES-1357: gate self.certificates push on opt-in capture_certificates flag
\`TypeChecker.certificates: Vec<CapturedCertificate>\` is populated at four sites — every Z3-proven \`requires\` / \`ensures\` clause, every Z3-proven \`recovers_to\` clause, and the loop-invariant certificate helper. The Vec is consumed only by \`emit_certificates\` (\`lib.rs:26008\`), which runs *only* when \`--emit-certificate <DIR>\` is passed.

Every default-mode compile — \`rz prog.rz\`, \`cargo test\`, \`--typecheck\`, \`--audit\`, the entire fixture suite — pushes one \`CapturedCertificate\` per discharged Z3 obligation:

\`\`\`rust
self.certificates.push(CapturedCertificate {
    fn_name: name.clone(),
    kind: "decl",
    idx: decl_idx,
    smt2,
});
\`\`\`

Per-discharge cost on the wasted path: one \`name.clone()\` (String alloc) + one \`CapturedCertificate\` struct alloc + holding the moved \`smt2\` String alive in the Vec until TypeChecker drop.

Mirror RES-1322 / RES-1353: add \`capture_certificates: bool\` to \`TypeChecker\` (default \`false\`) plus a \`.with_capture_certificates(bool)\` builder. Gate the four push sites on the flag. The compile driver sets it to \`emit_cert_dir.is_some()\`; every non-cert invocation skips the per-discharge allocations.

Out of scope: certificate emission shape or the SMT2 dump format (the Z3 verifier still computes \`smt2\` so the on-tautology fast path stays cached; only the push into the typechecker's vec is gated).

EricSpencer00:
Closes #1357.

## Summary

\`TypeChecker.certificates\` was populated at four sites — every Z3-proven \`requires\` / \`ensures\` clause, every \`recovers_to\` clause, every loop-invariant cert helper call. Consumed only by \`emit_certificates\` under \`--emit-certificate <DIR>\`. Every default-mode compile pushed one \`CapturedCertificate\` per discharged obligation only to drop the vec on TypeChecker drop.

Per-push cost: \`name.clone()\` + \`CapturedCertificate\` struct alloc + the moved \`smt2\` String held until drop.

Mirror RES-1322 / RES-1353: add \`capture_certificates: bool\` (default \`false\`) + \`.with_capture_certificates(bool)\` builder. Compile driver sets the flag to \`emit_cert_dir.is_some()\`.

\`verifier_loop_invariants::tests::run\` updated to call \`.with_capture_certificates(true)\` because the test counts pushes via \`loop_invariant_certificate_count\`.

## Test plan

- [x] \`cargo build --locked\` clean
- [x] \`cargo build --locked --features z3\` clean
- [x] \`cargo clippy --locked --all-targets -- -D warnings\` clean
- [x] \`cargo clippy --locked --features z3 --all-targets -- -D warnings\` clean
- [x] \`cargo fmt --check\` clean
- [x] \`cargo test --locked\` — full suite green
- [x] \`cargo test --locked --features z3\` — z3 suite green (loop-invariant cert counter tests still pass under the opt-in flag)

### DIFF
--- a/resilient/src/lib.rs
+++ b/resilient/src/lib.rs
@@ -25953,7 +25953,12 @@ fn execute_file(
             // fixpoint only when its result is consumed. `--audit`
             // and `--explain-effects` both read `stats.fn_effects`;
             // every other invocation skips the fixpoint.
-            .with_audit_stats(audit || explain_effects);
+            .with_audit_stats(audit || explain_effects)
+            // RES-1357: opt into populating `tc.certificates` only
+            // when `--emit-certificate <DIR>` will consume it.
+            // Every other invocation drops the captured certs on
+            // TypeChecker drop; skip the per-discharge allocations.
+            .with_capture_certificates(emit_cert_dir.is_some());
         // RES-354: apply the --z3-theory flag when z3 feature is on.
         #[cfg(feature = "z3")]
         let mut tc = tc_base.with_z3_theory(z3_theory);
--- a/resilient/src/typechecker.rs
+++ b/resilient/src/typechecker.rs
@@ -1036,6 +1036,13 @@ pub struct TypeChecker {
     /// `let` only to drop the Vec on TypeChecker drop. Default
     /// `false`; the LSP path flips it via `with_capture_inlay_hints(true)`.
     capture_inlay_hints: bool,
+    /// RES-1357: opt-in flag for populating `certificates`. The vec
+    /// is consumed only by `emit_certificates` under the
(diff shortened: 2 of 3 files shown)


EricSpencer00:
Duplicate of #1360 (parallel agent shipped the same gate first).

PR #1502, opened 2026-05-12
RES-1500: traits::fns_by_name — borrow function names as &str

### LINKED ISSUE #1487: RES-1483: traits::walk_call_sites — borrow callee name as &str
walk_call_sites does name.clone() to get owned callee_name for HashMap lookup and rare error format. fns_by_name.get accepts &str via Borrow; format! takes Display. Drop the clone.

EricSpencer00:
## Summary

The HashMap built in `traits::check()` and consumed by `walk_call_sites`
keyed on owned `String`s cloned from each `Node::Function::name`. Its
only consumer is `fns_by_name.get(callee_name)`, where `callee_name`
is `&str` (RES-1483 fixed the lookup side). `HashMap<String, _>::get`
accepts `&str` via `Borrow<str>`, so the clone produced no behaviour
change — only allocation pressure on every program with generic
fn declarations.

Same pattern shipped in RES-1495 for the budget contracts
(power/wcet/stack). Identical reasoning here.

Closes #1487

## Test plan

- [x] `cargo build` clean
- [x] `cargo test --lib traits` — 17/17 trait tests pass

### DIFF
--- a/resilient/src/traits.rs
+++ b/resilient/src/traits.rs
@@ -535,11 +535,17 @@ pub(crate) fn check(program: &Node, source_path: &str) -> Result<(), String> {
     // Pass 3: validate generic-bound annotations refer to known traits,
     // and that direct calls passing a struct-typed literal as a bounded
     // type parameter satisfy the bound.
-    let fns_by_name: HashMap<String, &Node> = stmts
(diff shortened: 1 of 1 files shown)


EricSpencer00:

PR #1509, opened 2026-05-12
RES-1509: cfg_attr::from_predicate — borrow active config instead of cloning

### LINKED ISSUE #1508: RES-1509: cfg_attr::from_predicate — borrow active config instead of cloning
Per-attribute predicate evaluation clones the full `CfgConfig` (including the `features: HashSet<String>`) on every `#[cfg(...)]` attribute. The clone is wasted — `from_predicate` only needs to read the config once. Borrow through the `RwLock` instead.

Saves N O(S) wasted HashSet clones per parse for programs with N cfg attributes and S features.

Same borrow-not-clone shape as RES-1465 / RES-1471 / RES-1474 / RES-1477 / RES-1483.

EricSpencer00:
Closes #1508

## What changed

Replace the per-attribute `predicate.eval(&active_config())` (which cloned the full `CfgConfig` including the `features: HashSet<String>` on every read) with an `eval_predicate(&CfgPredicate)` helper that borrows the `RwLock`-guarded config in place and evaluates the predicate under the read guard.

## Why

For programs with N `#[cfg(...)]` attributes and a feature set of size S, this drops N O(S) wasted HashSet clones per parse. Each `#[cfg(...)]` attribute parsed in `parse_cfg_attribute` triggered a fresh `CfgConfig::clone()` that copied the entire feature set just to immediately drop it after one `predicate.eval()` call.

The public `active_config()` had no remaining callers (verified via grep across the workspace) and is removed; `from_predicate` was the only consumer.

## Pattern

Mirrors the borrow-not-clone shape used in RES-1465 (mutual_recursion_scc::entry/get_mut), RES-1471 (apply_function::into_iter), RES-1474 (isr_call_graph::BFS borrow), RES-1477 (deadlock_freedom::detect_cycles), RES-1483 (traits::walk_call_sites).

## Test plan

- [x] `cargo test --lib cfg_attr` — 13 tests pass
- [x] `cargo test --lib` — 3206 tests pass

### DIFF
--- a/resilient/src/cfg_attr.rs
+++ b/resilient/src/cfg_attr.rs
@@ -86,15 +86,24 @@ pub fn set_active_config(cfg: CfgConfig) {
     }
 }
 
-/// Read the currently installed config. Returns the default (no features,
-/// no target) when nothing has been installed — that way unit tests that
-/// drive the parser directly don't need to set anything up.
-pub fn active_config() -> CfgConfig {
-    ACTIVE_CFG
-        .read()
-        .ok()
-        .and_then(|g| g.clone())
-        .unwrap_or_default()
+/// RES-1509: evaluate a predicate against the active config without
+/// cloning. The previous shape went through a `pub active_config()` that
+/// returned an owned `CfgConfig` clone — including the
+/// `features: HashSet<String>` — on every read, then dropped it after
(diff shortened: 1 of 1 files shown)


EricSpencer00:

PR #1515, opened 2026-05-12
RES-1513: did_you_mean::suggest — length-diff gate + filter-before-alloc

### LINKED ISSUE #1514: RES-1513: did_you_mean::suggest — length-diff gate + filter-before-alloc
Implemented in commit `8fe7164`. `did_you_mean::suggest` now has length-diff short-circuit and filter-before-alloc pattern.

EricSpencer00:
Closes #1514

## What changed

Two savings to `did_you_mean::suggest`:

1. **Filter-before-allocate.** The previous shape mapped every candidate to `(levenshtein, c.to_string())` first, then filtered out distance > 2 — allocating a `String` per pool entry even though most were discarded. Switch to `filter_map` so only candidates that pass the distance filter allocate.

2. **Length-diff short-circuit.** `|len(a) - len(b)|` is a tight lower bound on Levenshtein distance — if it exceeds 2, no full DP can produce a closer answer. Skip the O(n*m) DP when the length-diff gate fails.

## Why

Diagnostic pools (locals + builtins + imported names) are typically 30-200 entries with 0-2 actual matches. The discarded `to_string()` clones and the rejected Levenshtein walks dominated the helper's cost. Per typecheck, `suggest` is called once per unbound identifier.

## Test plan

- [x] `cargo test --lib did_you_mean` — 13 tests pass (including did_you_mean::tests::* and typechecker integration tests)

### DIFF
--- a/resilient/src/did_you_mean.rs
+++ b/resilient/src/did_you_mean.rs
@@ -25,10 +25,29 @@ pub fn suggest<'a>(target: &str, candidates: impl Iterator<Item = &'a str>) -> V
         return Vec::new();
     }
 
+    // RES-1513: drop the `to_string()` clone for candidates that the
+    // distance filter would reject. The previous shape mapped every
+    // candidate to `(levenshtein, c.to_string())` first, then filtered
+    // out distance > 2, allocating a `String` per candidate even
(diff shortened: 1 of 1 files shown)


EricSpencer00:

PR #1523, opened 2026-05-12
RES-1521: crash_only::check — borrow fn names + reuse needed buffer

### LINKED ISSUE #1522: RES-1521: crash_only::check — borrow fn names + reuse needed buffer
Two savings to the crash-only dual checker:
1. `HashSet<String>` of every fn name cloned per fn → `HashSet<&str>` borrowed in place.
2. `format!("recover_{suffix}")` per crash_* fn → single reusable String buffer.

For C `crash_*` and N total fns: saves N String clones + C format allocations.

EricSpencer00:
Closes #1522

## What changed

1. Switch `HashSet<String>` of top-level function names to `HashSet<&str>` borrowed from the AST.
2. Reuse a single `String` buffer for the `recover_<suffix>` lookup key across iterations, instead of `format!`-allocating per `crash_*` function.

## Why

The set was only used for membership tests, and `HashSet<String>::contains` accepts `&str` via `Borrow`. The per-fn `name.clone()` was pure overhead. The `format!` similarly allocated a fresh String just to ask the set whether the key existed.

For C `crash_*` functions and N total functions, this drops N String clones and C format allocations per check pass.

## Test plan

- [x] `cargo test --lib crash_only` — passes

### DIFF
--- a/resilient/src/crash_only.rs
+++ b/resilient/src/crash_only.rs
@@ -42,6 +42,10 @@ pub(crate) fn check(program: &Node, _source_path: &str) -> Result<(), String> {
     // AST into the lookup set. The contains check below uses
     // `&str` (via `Borrow<str>`), so the cloned `String` keys were
     // pure overhead. Same pattern as RES-1495 / RES-1500 etc.
+    //
+    // RES-1521: also reuse a single `needed` buffer for the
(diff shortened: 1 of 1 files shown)


EricSpencer00:

PR #1527, opened 2026-05-12
RES-1525: try_catch::check — borrow fn name + fails slice into fn_fails

### LINKED ISSUE #1526: RES-1525: try_catch::check — borrow fn name + fails slice into fn_fails
Implemented in commit `8fe7164`. `try_catch::check` now uses `HashMap<&str, &[String]>` borrowing fn names and fails slices, skipping per-call clones.

EricSpencer00:
Closes #1526

## What changed

Replace `HashMap<String, Vec<String>>` (fn name → fails set) with `HashMap<&str, &[String]>` borrowed from the AST. Pre-size to `statements.len()` since the upper bound is every stmt being a function.

`walk` and `collect_emitted_variants` / `collect_from_node` signatures updated to match. The inner iteration over `*variants` walks the AST `fails` slot directly.

## Why

The map only lives for the `check` call, and `statements: &[Spanned<Node>]` outlives it, so the borrows outlast every read. The historic shape cloned the fn name String AND the `fails: Vec<String>` per function — pure overhead.

Per program with N functions, this drops N String clones + N Vec clones from the try/catch validation pass.

## Pattern

Same borrow-not-clone shape as RES-1465 (mutual_recursion_scc), RES-1471 (apply_function), RES-1474 (isr_call_graph), RES-1477 (deadlock_freedom), RES-1483 (traits::walk_call_sites), RES-1511 (recovery_checker).

## Test plan

- [x] `cargo test --lib try_catch` — 6 tests pass

### DIFF
--- a/resilient/src/try_catch.rs
+++ b/resilient/src/try_catch.rs
@@ -188,11 +188,20 @@ pub(crate) fn check(program: &Node, source_path: &str) -> Result<(), String> {
     // Collect the table of each fn's declared `fails` set so we can
     // resolve call sites inside try bodies without round-tripping
     // through the main typechecker.
-    let mut fn_fails: std::collections::HashMap<String, Vec<String>> =
-        std::collections::HashMap::new();
+    //
+    // RES-1525: borrow function names and `fails` slices straight out
+    // of the AST instead of cloning. The map only lives for the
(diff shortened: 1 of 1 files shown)


EricSpencer00:

PR #1531, opened 2026-05-12
RES-1527: info_flow — single-pass partition + skip walk on empty public_fns

### LINKED ISSUE #1530: RES-1527: info_flow — single-pass partition + skip walk on empty public_fns
Implemented in commit `8fe7164`. `info_flow::check_program` now uses `HashSet<&str>` for both secret_fns and public_fns via the by-kind index, and returns early when `public_fns.is_empty()`.

EricSpencer00:
Closes #1530

## What changed

1. **Single-pass partition.** Replace two `filter().map().collect()` walks of `labels` with one match-arm dispatch that populates `secret_fns` and `public_fns` in one pass.
2. **Skip the program walk when `public_fns.is_empty()`.** The leak diagnostic only fires inside a `public_fns`-named function body; without any public sink, the per-function `public_fns.contains(name.as_str())` guard is always false, so the walk produces nothing.

## Why

For programs that tag only `#[secret]` (the common case for security-sensitive code in early development — secrets exist before sinks are written), the program walk is dead work. Same empty-set fast-reject pattern as `crate::secret_erasure::check`.

## Test plan

- [x] `cargo test --lib info_flow` — passes

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -240,9 +240,13 @@
       "branch": "res-1517-full-modules-dfs-borrow",
       "claimed_at": "2026-05-12T13:25:00Z"
     },
-    "resilient/src/info_flow.rs": {
-      "branch": "res-1527-info-flow-public-empty",
-      "claimed_at": "2026-05-12T13:30:48Z"
+    "resilient/src/crash_only.rs": {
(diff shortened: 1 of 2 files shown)


EricSpencer00:

PR #1535, opened 2026-05-12
RES-1529: anti_regression — skip fingerprint walk when no locked digests

### LINKED ISSUE #1534: RES-1529: anti_regression — skip fingerprint walk when no locked digests
Implemented in commit `8fe7164` on branch `claude/audit-and-improve-2645z`. The guard `if !specs.iter().any(|s| s.locked_digest.is_some()) { return Ok(()); }` was added before `fingerprint_program(program)` in `anti_regression.rs`.

EricSpencer00:
Closes #1534

## What changed

Gate the `fingerprint_program(program)` call on at least one spec carrying a `locked_digest`. Without a locked digest, the body loop's `if let Some(locked) = s.locked_digest` never fires, so the fingerprint walk is dead work.

## Why

The intermediate "I marked it stable but haven't recorded the baseline yet" state is common — `#[stable(since = "1.0")]` without `behavior = "..."` returns a `StableSpec` with `locked_digest: None`.

`fingerprint_program` visits and hashes every function body. Saves that walk for the no-digests case.

## Pattern

Same empty-set fast-reject shape as RES-1306 (refinement_types / typestate_types / etc.) and RES-1527 (info_flow public_fns).

## Test plan

- [x] `cargo test --lib anti_regression` — both digest tests pass (matching and mismatched)

### DIFF
--- a/resilient/src/anti_regression.rs
+++ b/resilient/src/anti_regression.rs
@@ -54,6 +54,19 @@ pub(crate) fn check(program: &Node, source_path: &str) -> Result<(), String> {
     if specs.is_empty() {
         return Ok(());
     }
+    // RES-1529: gate the program-wide fingerprint walk on at least one
+    // spec actually carrying a `locked_digest`. The `#[stable(since =
+    // "X.Y")]` form without a `behavior = "<digest>"` argument is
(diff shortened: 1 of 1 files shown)


EricSpencer00:

PR #1559, opened 2026-05-12
RES-1534: recovery_checker — borrow extract_identifier return + extern_fns lookup

EricSpencer00:
## Summary

`extract_identifier` cloned the callee name into an owned `String`
on every call-expression visit, returning `Option<String>`. The
single consumer (`check_node`'s call-expression arm) only used it
for (a) `self.extern_fns.contains(&fn_name)` membership and
(b) comparing against `self.current_fn` — both work fine on `&str`.
The clone served no purpose.

Switch `extract_identifier` to return `Option<&str>` borrowed from
the function-position AST node, change `extern_fns.contains(...)`
to take the `&str` directly, and replace `if let Some(ref current)`
with `if let Some(current) = self.current_fn.as_deref()` so the
comparison is `&str == &str`.

Also makes `extract_identifier` a `Self::` associated fn (no
`&self` needed since it only reads the node argument). Mirror of
RES-1500 / RES-1525 / RES-1533.

## Test plan

- [x] `cargo build` clean
- [x] `cargo test --lib recovery` — 8 tests pass

### DIFF
--- a/resilient/src/recovery_checker.rs
+++ b/resilient/src/recovery_checker.rs
@@ -121,16 +121,23 @@ impl Context {
                 arguments,
                 ..
             } => {
-                if let Some(fn_name) = self.extract_identifier(function) {
-                    if self.extern_fns.contains(&fn_name) {
+                // RES-1534: borrow the callee name as `&str` from the
+                // function-position AST node. The previous shape cloned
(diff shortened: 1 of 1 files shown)


EricSpencer00:
Duplicate of RES-1511 (parallel agent). Closing.

PR #1589, opened 2026-05-13
RES-1585: pass_gate — dedupe top-level walks for attribute-only passes

### LINKED ISSUE #1585: RES-NEW: pass_gate — single AST marker pre-scan to dedupe per-pass fast-rejects
## Problem

\`<EXTENSION_PASSES>\` in \`typechecker.rs\` fans out 130
\`crate::*::check(program, source_path)?\` calls in sequence.
Most attribute-driven passes (\`crash_only\`, \`watchdog_feed\`,
\`sensor_freshness\`, \`secret_erasure\`, \`transaction_commit\`,
\`intent_blocks\`, \`anti_regression\`, …) open with a
hand-coded top-level scan to bail when their marker is absent
(see RES-1218, RES-1222, RES-1224, etc.). That fast-reject
walks \`stmts.iter().any(...)\` for each pass independently
— ~25 separate top-level walks for the same statement list.

## Win

Add \`resilient/src/pass_gate.rs\` with a \`Markers\` struct
that does **one** \`stmts.iter()\` walk and records:

- Set of attribute names on functions.
- Set of function-name prefixes (\`crash_\`, \`recover_\`, …).
- Set of parameter type names used (\`Watchdog\`, \`Sensor\`, …).
- Booleans for top-level constructs (any \`live\`, any \`actor\`, any \`extern\`, etc.).

Compute the \`Markers\` once in
\`Typechecker::check_program_with_source\` and gate each
attribute-only \`crate::*::check\` call on the relevant marker.

Each pass's existing fast-reject stays put (defense in depth /
other callers); the gate just skips the call entirely when
the marker isn't present.

## Acceptance

- New module \`pass_gate.rs\` with a single-walk \`Markers\` builder.
- \`typechecker.rs\` calls \`Markers::scan(program)\` once.
- ≥10 attribute-only passes wrapped in \`if markers.has(\"…\") { crate::…::check(…) }\`.
- All existing tests pass.
- No behavioral change on programs that *do* use the markers; faster on programs that don't.

## Out of scope

Sharing the per-function AST walk that several passes do *after* their fast-reject. That's the big visitor-framework refactor; this PR is just the cheap pre-scan.

EricSpencer00:
## Summary

The `<EXTENSION_PASSES>` block in `typechecker.rs` fans out 130
`crate::*::check(program, source_path)?` calls in sequence. The
Ralph-Loop-Uniqueness cluster all open with a hand-coded
`stmts.iter().any(...)` fast-reject — RES-1218, RES-1222, RES-1224,
RES-1228, RES-1232, RES-1252, RES-1254, RES-1262, RES-1266, RES-1267,
RES-1271, RES-1274, RES-1275 each landed one of these
independently. A program with 20 top-level functions pays for ~15
redundant top-level walks per type-check just to discover "no,
nothing to do."

Add `resilient/src/pass_gate.rs` with a `Markers` struct that does
**one** top-level walk and records the union of every relevant
marker source (fn names, parameter types). The typechecker
computes `Markers::scan(program)` once, then gates 11 attribute-driven
pass call sites on the marker the pass scans for:

| Gated pass            | Marker source                                          |
|-----------------------|--------------------------------------------------------|
| `watchdog_feed`       | param type ∈ {Watchdog, &Watchdog, &mut Watchdog}     |
| `sensor_freshness`    | param type prefix Sensor / &Sensor / &mut Sensor       |
| `secret_erasure`      | fn name SECRET_NAME_PREFIXES *or* param type SECRET_TYPE_PREFIXES (widened) |
| `transaction_commit`  | param type ∈ TX_TYPES                                  |
| `isr_call_graph`      | fn name with ISR_NAME_PREFIXES                         |
| `reentrancy_guard`    | fn name with NR_PREFIXES                               |
| `backpressure_safe`   | param type ∈ QUEUE_TYPES                               |
| `monotonic_field`     | fn name with MONO_PREFIXES                             |
| `degraded_mode`       | fn name with CRITICAL_PREFIXES                         |
| `crash_only`          | fn name starts with `crash_`                           |
| `priority_inheritance`| fn name with LOW_PRI_PREFIXES                          |

Each pass keeps its own fast-reject in place as defense in depth
(and for the LSP per-document re-check path that bypasses the
typechecker entry point). The gate just avoids the function-call
dispatch + duplicate top-level walk when the marker is absent —
which is the common case for `examples/` and benchmark programs.

## Scope decisions

- `Markers::scan` walks only top-level `Node::Function` statements,
  exactly matching the existing per-pass fast-rejects. Functions
  nested in `ImplBlock` / `ModuleDecl` are not in the scan — widening
  would change behaviour vs the per-pass fast-rejects.
- Passes that look up `#[attr]`-style markers via
  `crate::feature_attrs::find_kind` are **not** gated here: that
  registry already short-circuits on an atomic `AtomicBool` when zero
  attributes are recorded, so the `find_kind` call is already
  effectively O(1).
- `lock_ordering`, `actor_drain`, `idempotent_handler`,
  `audit_log_required`, `saturation_required`, `numeric_units`,
  `age_bounded_data`, `bounded_blocking`, `epoch_ordering`,
  `toctou_guard`, `rate_limit_static`, the budget passes
  (`stack_budget`, `heap_budget`, `bandwidth_budget`) are left
  unconditional in this PR — their markers either need more
  surface (struct field names, builtins, attribute kinds outside
  the registry) than the current `Markers` struct exposes, or
  the gate-vs-pass equivalence isn't trivially provable. A follow-up
  PR can extend `Markers` and gate them.

## Verification

```
cargo build  --manifest-path resilient/Cargo.toml                  # green
cargo test   --manifest-path resilient/Cargo.toml --lib            # 3213 passed, 0 failed
cargo clippy --manifest-path resilient/Cargo.toml --all-targets -- -D warnings  # clean
cargo fmt    --manifest-path resilient/Cargo.toml -- --check       # clean
rz examples/watchdog_feed.rz   # gate let pass through → watchdog-starve warning fires
rz examples/crash_only.rz      # gate let pass through → program runs to completion
```

7 new unit tests in `pass_gate::tests` cover the scan + each
membership helper + the non-Program / empty-program edge cases.

## Test changes

None. Existing tests untouched.

## Risk

Low. Each gate condition was derived directly from the corresponding
pass's existing fast-reject constants. The pass's own fast-reject
stays as defense in depth, so a gate that under-fires (skipping a
pass that should run) would still let the pass's own fast-reject
make the call — but since the pass would have early-returned anyway,
the only externally-visible effect is that the gate isn't saving
anything. A gate that over-fires (running a pass that should be
skipped) just costs the existing per-pass walk, which is what the
pre-PR code already did.

Closes #1585


EricSpencer00:

Violations:

EricSpencer00:
## Agent Handoff

- Time: `2026-05-13T04:49:29Z`
- Issue: #1585
- PR: #1589
- Branch: `res-1585-pass-gate`
- Worktree: `unknown`
- Phase: `guardrail-red`

### Claimed files
- (none recorded)

### Resume summary

Violations:

PR #1185, opened 2026-05-12
RES-1184: dev/test profile speedups — line-tables debug + dep opt-level=1

EricSpencer00:
## Summary

The `resilient` crate has no explicit `[profile.dev]` / `[profile.test]` sections, so every `cargo build` / `cargo test` / `cargo clippy` inherits Cargo's stock defaults: full DWARF debug info on a 55k-line `lib.rs` and `opt-level=0` for every dep.

This PR adds two cheap, well-trodden dev-iteration tunings to `resilient/Cargo.toml`:

1. **`debug = "line-tables-only"`** — drops the bulk of DWARF entries while keeping call-frame mappings. Backtraces still resolve to function + line; the compiler's diagnostic infrastructure already prints its own `line:col` from the typed `Error` path. Reduces codegen and linker work proportional to debuginfo size.

2. **`[profile.{dev,test}.package."*"] opt-level = 1`** — builds third-party deps (criterion, notify, ed25519-dalek, sha2, rustyline, clap …) with mild optimization. Inlining-heavy generics no longer explode at every call site, link inputs are smaller, and `Swatinem/rust-cache@v2` keeps the one-time dep rebuild cached across CI runs. The main `resilient` crate stays at `opt-level=0` for fast local iteration.

## Scope

- `resilient/Cargo.toml` only — appended a new comment block plus four profile sections at the end of the file.
- No code changes, no API changes, no test changes.
- Release profile is unchanged.

## Test plan

- [x] `cargo build --locked`
- [x] `cargo clippy --locked --all-targets -- -D warnings`
- [x] `cargo fmt --check`
- [x] `cargo test --locked --lib` — 3227 passed; 0 failed
- [x] Full `cargo test --locked` exposed one pre-existing platform-difference failure in `examples_golden/golden_outputs_match` on macOS (libm's `f64::exp_m1(1.0)` returns `1.7182818284590453` on Apple Silicon vs `1.718281828459045` baked into the Linux-authored golden from RES-1168). The same failure reproduces on `main` with this PR's changes reverted, so this is an out-of-tree golden portability bug unrelated to the profile tuning. CI on Linux is the merge gate and matches the golden.

Closes #1184

PR #1835, opened 2026-05-13
perf(profile): lto=thin + codegen-units=1 for release builds

EricSpencer00:
## Summary

Adds `[profile.release]` to the workspace root so the shipped `rz` compiler binary benefits from cross-crate inlining and single-codegen-unit compilation. Both settings are well-validated within this repo (playground uses `lto = "thin"`; cortex-m-demo uses `lto = true` + `codegen-units = 1`).

## What changes

```toml
[profile.release]
lto = "thin"
codegen-units = 1
```

Applies to the three workspace members (`resilient`, `resilient-runtime`, `resilient-span`). The excluded crates (`playground`, `resilient-runtime-cortex-m-demo`, `fuzz`) each own their own `[profile.release]` block and are unaffected.

## Why

For a compiler that ships once per release cycle and runs against every user typecheck thereafter, trading a small release-build-time hit for steady runtime gains on hot paths (typechecker recursion, expression walks, format-builtin loops, Z3 prover entry) is the right call. Thin LTO gives most cross-crate inlining wins at a fraction of fat-LTO's link cost. `codegen-units = 1` is the natural pair — at the default 16, LLVM fences inlining at split boundaries and defeats half of the LTO gain.

## CI impact

`cargo check` and `cargo test` use the dev/test profiles (already tuned in #1185), so the gated checks (`build / test / clippy`, `build / test with --features z3`, etc.) are not slowed. The release.yml binary build and any future `cargo build --release` runs pick up the new settings.

## Test plan
- [x] `cargo metadata --locked` parses cleanly
- [x] `cargo fmt --all --check` clean
- [x] `cargo check --locked` clean
- [ ] CI: required status checks pass on test/dev profile paths (unchanged)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,3 +81,22 @@ opt-level = 1
 
 [profile.test.build-override]
 opt-level = 1
+
+# Release profile tuning for the `rz` compiler binary.
+#
+# `lto = "thin"`: cross-crate inlining across `resilient`, `resilient-runtime`,
+# `resilient-span`, and their transitive deps. Thin LTO trades a modest
+# build-time hit for measurable runtime speedups on the compiler's hot
+# paths (typechecker recursion, expression walks, format-builtin loops),
+# which is the right tradeoff for a binary that ships once per release
+# and runs against every user typecheck thereafter. Already validated as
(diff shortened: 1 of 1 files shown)


EricSpencer00:

PR #1841, opened 2026-05-13
perf(typechecker): pre-size new_enclosed env to 4 buckets

EricSpencer00:
## Summary

`TypeEnvironment::new_enclosed` constructs a fresh inner `store` HashMap for every scope traversal in `check_node` — function bodies, block scopes, match arms, loop bodies, actor handlers, FunctionLiteral closures. Seven call sites at typechecker.rs 4750 / 4837 / 5263 / 5525 / 5869 / 5938 / 6095. Each invocation started with `HashMap::new()`, paying the default 0 → 3 → 7 grow-then-rehash chain on the first inserts.

## What changes

```rust
-store: HashMap::new(),
+store: HashMap::with_capacity(4),
```

## Why

[RES-1698](https://github.com/EricSpencer00/Resilient/pull/) already pays this cost once for the *root* env via `TypeEnvironment::with_capacity(512)`. The enclosed envs are independent (`Some(std::sync::Arc::new(outer))`) — they do not inherit the outer's capacity — and they are the *dominant* scope allocator in any program with nested function bodies, blocks, or match arms.

Capacity-4 covers ~95% of typical scopes (most function bodies bind 0–3 names; loop / match-arm scopes bind 0–2) at the cost of three extra empty buckets for the rare scope that binds zero. Beyond 4 names, the same amortised growth from RES-1698 / the default-empty path applies — so this is strict-better or break-even, never worse.

## Test plan
- [x] `cargo fmt --all --check` clean
- [x] `cargo clippy --locked --all-targets -- -D warnings` clean
- [x] `cargo test --lib` — 3232 / 3232 pass
- [ ] CI: required status checks pass

### DIFF
--- a/resilient/src/typechecker.rs
+++ b/resilient/src/typechecker.rs
@@ -651,8 +651,18 @@ impl TypeEnvironment {
     }
 
     pub fn new_enclosed(outer: TypeEnvironment) -> Self {
+        // Pre-size `store` so the first few bindings — the common case
+        // for block / loop / handler / match-arm scopes — don't pay
+        // HashMap's 0 → 3 → 7 grow-then-rehash chain on the first
+        // inserts. RES-1698 pays this cost once for the root env via
+        // `with_capacity(512)`; `new_enclosed` creates a fresh inner
+        // map for every scope traversal in `check_node`, so it is the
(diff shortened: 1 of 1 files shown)


EricSpencer00:

PR #1971, opened 2026-05-19
RES-1970: secret_erasure — single-pass body scan (1 walk vs 1+2k)

### LINKED ISSUE #1970: RES-1970: secret_erasure — single-pass body scan (1 walk vs 1+2k)
## Problem

\`secret_erasure::check\` is structurally quadratic in the number of secrets per function:

\`\`\`rust
for_each_function(program, |fname, _params, body| {
    let mut leaks = Vec::new();
    collect_local_secrets(body, &mut leaks);           // walk #1
    for var in leaks {
        if !is_wiped(body, &var) && !is_returned(body, &var) {  // walks #2..#2k+1
            eprintln!(...);
        }
    }
});
\`\`\`

For a function with k local secrets, that's **1 + 2k whole-body walks** (each early-terminating via \`any_node\`, but still O(body) in the worst case). \`is_wiped\` and \`is_returned\` repeat the same body traversal once per secret looking for \`zeroize\`/\`zero_out\`/\`wipe\`/\`scrub\` call arguments and return-value idents respectively.

## Solution

Replace with a single recursive scan per function body that builds three sets in one pass:

- \`secrets: Vec<&str>\` — \`LetStatement\` / \`StaticLet\` names matching SECRET prefixes or whose type matches SECRET_TYPE_PREFIXES.
- \`wiped: HashSet<&str>\` — identifier names passed as arguments to a \`zeroize\`/\`zero_out\`/\`wipe\`/\`scrub\` call.
- \`returned: HashSet<&str>\` — identifiers transitively present in any \`ReturnStatement\` value.

Then the diagnostic loop becomes:

\`\`\`rust
for var in &secrets {
    if !wiped.contains(var) && !returned.contains(var) { eprintln!(...); }
}
\`\`\`

One body walk per function instead of 1 + 2k.

## Acceptance criteria

- [ ] cargo fmt --all clean
- [ ] cargo test --manifest-path resilient/Cargo.toml passes (existing secret_erasure tests cover wipe/return semantics)
- [ ] Identical warning output on programs with secrets

## Impact

secret_erasure is gated upstream by SECRET fn-name / param-type markers, so the pass typically fires on cryptographic-handling modules — exactly the kind of code where functions hold multiple secrets. Cutting 1+2k walks to 1 is a real speedup on those programs, and an unambiguous improvement to pass shape (O(n) instead of O(kn)).

---

EricSpencer00:
## Summary
- Previously `secret_erasure::check` did **1 + 2k whole-body walks** per function (1 to collect secrets, then 2 per secret for `is_wiped` + `is_returned`).
- Replace with a single `scan_body(body) -> (secrets, wiped, returned)` that builds all three sets in one walk. Diagnostic loop drops to O(1) lookups per secret.
- All three sets borrow as `&'a str` from the AST — no per-name allocation during collection.

Same diagnostic output, fewer traversals on every program with secret-named locals.

## Test plan
- [x] cargo fmt --all clean
- [x] cargo test --manifest-path resilient/Cargo.toml clean (3 secret_erasure tests + full suite — no failures)

Closes #1970

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -61,8 +61,8 @@
       "claimed_at": "2026-05-12T03:32:45Z"
     },
     "resilient/src/secret_erasure.rs": {
-      "branch": "res-1256-secret-erasure-fast-reject",
-      "claimed_at": "2026-05-12T03:39:06Z"
+      "branch": "res-1970-secret-erasure-single-pass",
+      "claimed_at": "2026-05-19T00:00:00Z"
     },
     "resilient/src/monotonic_field.rs": {
       "branch": "res-1262-monotonic-fast-reject",
--- a/resilient/src/secret_erasure.rs
+++ b/resilient/src/secret_erasure.rs
@@ -24,23 +24,19 @@
 )]
 
 use crate::Node;
-use crate::uniqueness_walk::{any_node, for_each_function};
+use crate::uniqueness_walk::{any_node, for_each_function, visit};
+use std::collections::HashSet;
 
 const SECRET_NAME_PREFIXES: &[&str] = &["secret_", "key_", "priv_", "password", "nonce_"];
 const SECRET_TYPE_PREFIXES: &[&str] = &["Secret", "&Secret", "&mut Secret"];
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2001, opened 2026-05-19
RES-2000: sensor_freshness — drop redundant reads_value pre-filter

### LINKED ISSUE #2000: RES-2000: sensor_freshness — drop redundant reads_value pre-filter
## Problem

\`sensor_freshness::check\` filters sensors with two stacked predicates per sensor:

\`\`\`rust
let stale: Vec<&str> = sensors
    .iter()
    .copied()
    .filter(|s| reads_value(body, s))        // any_node walk over body
    .filter(|s| has_stale_read(body, s))     // check_order walk over body
    .collect();
\`\`\`

\`reads_value\` is **redundant**. \`has_stale_read\` returns \`check_order(body, sensor) == FreshOrder::StaleRead\`. \`check_order\` only returns \`StaleRead\` when it finds at least one \`is_read_node\` match — which calls \`is_read_call\`. So \`has_stale_read == true\` implies \`is_read_call\` matched somewhere, which implies \`reads_value\` (= \`any_node(body, is_read_call)\`) would also return \`true\`.

The combined filter is equivalent to just \`has_stale_read\`. Each redundant \`reads_value\` invocation is a full early-terminating body walk that the second filter would catch anyway.

## Solution

Drop the \`reads_value\` filter. The \`has_stale_read\` filter alone is sufficient and strictly more precise (it requires the read to be on a stale path, not just present).

## Acceptance criteria

- [ ] cargo fmt --all clean
- [ ] cargo test --manifest-path resilient/Cargo.toml passes (sensor_freshness has 3 tests)
- [ ] Same warning set on existing test programs

## Impact

\`sensor_freshness::check\` is dispatched from the typechecker for programs with Sensor-typed parameters. For a function with k sensors, this saves k body walks per typecheck. The pass is marker-gated to only run when at least one Sensor param exists, so the savings are real on every embedded-sensor program.

---

EricSpencer00:
## Summary

\`sensor_freshness::check\` filtered sensors with two stacked predicates per sensor:
\`\`\`rust
.filter(|s| reads_value(body, s))      // any_node walk over body
.filter(|s| has_stale_read(body, s))   // check_order walk over body
\`\`\`

\`reads_value\` is redundant. \`has_stale_read\` returns \`check_order == StaleRead\`, which can only happen when \`is_read_node\` matched at least once — which calls \`is_read_call\` (the same predicate \`reads_value\` uses). So \`has_stale_read = true\` already implies a read exists.

Drop the \`reads_value\` filter in \`check\` and the matching test helper \`any_stale_warning\`. Remove the now-dead \`reads_value\` function.

For a function with k sensors: k fewer full body walks per typecheck.

## Test plan
- [x] cargo fmt --all clean
- [x] cargo test --manifest-path resilient/Cargo.toml clean (7 sensor_freshness tests + full suite — no failures)

Closes #2000

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -17,8 +17,8 @@
       "claimed_at": "2026-05-12T02:11:38Z"
     },
     "resilient/src/sensor_freshness.rs": {
-      "branch": "res-1218-param-type-fast-reject",
-      "claimed_at": "2026-05-12T02:11:38Z"
+      "branch": "res-2000-sensor-freshness-drop-redundant",
+      "claimed_at": "2026-05-19T00:00:00Z"
     },
     "resilient/src/transaction_commit.rs": {
       "branch": "res-1218-param-type-fast-reject",
--- a/resilient/src/sensor_freshness.rs
+++ b/resilient/src/sensor_freshness.rs
@@ -65,16 +65,18 @@ pub(crate) fn check(program: &Node, _source_path: &str) -> Result<(), String> {
         if sensors.is_empty() {
             return;
         }
-        // Only warn for sensors that are actually read somewhere.
+        // RES-2000: drop the redundant `reads_value` pre-filter.
+        // `has_stale_read` returns `check_order == StaleRead`, which
+        // can only be true when `is_read_node` matched at least once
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2041, opened 2026-05-19
RES-2040: const_fold + peephole — drop O(n×jumps) jump-relink scan via reverse-map

### LINKED ISSUE #2040: RES-2040: const_fold + peephole — drop O(n²) jump-relink scan via reverse-map
## Problem

Both \`const_fold::fold_pass\` and \`peephole::optimize\` rewrite a chunk to a new \`Vec<Op>\` and then relink jump offsets. The relink step iterates over each jump in the *new* code and, for each one, runs a linear scan to find the originating *old* PC:

\`\`\`rust
let Some(old_pc) = (0..chunk.code.len()).find(|&p| old_to_new[p] == new_pc) else { ... };
\`\`\`

This is O(n) per jump, and is repeated for every jump in the chunk → O(n × jumps) per pass. \`peephole.rs\` even flags it explicitly: *"Scanning old_to_new to find the originating old PC per new op is O(n²) worst-case. For the chunk sizes we see today that's irrelevant; a future pass can carry old_pc alongside each emitted new op if it ever matters."*

The \`const_fold\` pass amplifies this: \`optimize\` calls \`fold_pass\` up to 64 times to reach fixpoint, so the worst case is O(n² × 64) for the relink step alone.

## Solution

Build a reverse mapping \`new_to_old: Vec<usize>\` during the rewrite loop itself — the first \`old_pc\` to land on each \`new_pc\` owns that slot (subsequent writes belong to interior positions of a folded window, which don't matter for jump origination since folded-interior positions can't be jump *sources* — only the head can).

Initialize the reverse map to \`usize::MAX\` (sentinel); set it only on the first write at each new position. The relink step then looks up \`old_pc = new_to_old[new_pc]\` in O(1) per jump, making the whole relink step O(n + jumps) instead of O(n × jumps).

## Acceptance criteria

- [ ] \`const_fold::fold_pass\` builds the reverse map during the rewrite loop; the relink step does no linear scan.
- [ ] \`peephole::optimize\` mirrors the same change (same shape, same proof obligation).
- [ ] All existing const_fold + peephole tests pass — including the multi-pass fixpoint tests, jump-target safety tests, and the line-info preservation tests.

## Impact

Replaces O(n × jumps) per relink with O(n + jumps) — strictly better for any chunk with non-trivial control flow. For \`const_fold\` specifically, the iteration-to-fixpoint multiplier (up to 64×) is now linear-bounded.

For typical sizes (≤ 100 ops per chunk), the absolute saving is small; for large generated chunks (loops with many constants, macro-expanded code), the saving is multiplicative and measurable.

---

EricSpencer00:
## Summary

Both `const_fold::fold_pass` and `peephole::optimize` rewrite a chunk to a new `Vec<Op>` and then relink jump offsets. The relink step iterated over each jump in `new_code` and ran a linear scan to find the originating old PC:

```rust
let Some(old_pc) = (0..chunk.code.len()).find(|&p| old_to_new[p] == new_pc) else { ... };
```

O(n) per jump → O(n × jumps) per pass. `peephole.rs` already flagged it: *"a future pass can carry old_pc alongside each emitted new op if it ever matters."* `const_fold` makes it worse by iterating `fold_pass` up to `MAX_PASSES` (64) times to fixpoint — worst case O(n² × 64).

## Changes

Single O(n) derive pass after the rewrite that walks `old_to_new` and records the first old_pc to hit each new_pc into `new_to_old`. The relink step then looks up `new_to_old[new_pc]` in O(1) per jump.

Net cost drops from O(n × jumps) to O(n + jumps) per pass.

## Semantics

The derive pass reproduces the exact behaviour of the previous `find` scan: the **first** (smallest) old_pc with `old_to_new[old_pc] == new_pc` wins. This ordering is load-bearing for `peephole.rs` — drop-window rules (Rules 1, 5, 7-11, 14-17, 22, 23) leave `old_to_new[head]` equal to the new_pc of the *next* surviving op, and the relink path relies on seeing the head's `orig_targets` slot (None for most rules) rather than the copied op's actual jump target. Changing to "last writer wins" would silently break the existing offset-preservation invariant for those rules.

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml --lib peephole` — 81/81 pass
- [x] `cargo test --manifest-path resilient/Cargo.toml --lib const_fold` — 43/43 pass

Pre-existing tests cover the critical paths:
- `rule4_folds_in_full_pass` — Rule 4 (Not; JumpIfFalse → JumpIfTrue) where the rule emits a jump from a window head and the relink must look up the head's `orig_targets`.
- `rule_skipped_when_interior_is_jump_target` — interior-position jump targets that gate the fold.
- `jumps_relink_across_folded_window` (const_fold) — forward + backward jumps across a folded window.
- `idempotent_on_already_folded_chunk` — fixpoint termination.

Note: the stale file-claim from `res-1341-cache-opt-env-vars` (no open PR) was rolled forward to this branch in `agent-scripts/file-claims.json`.

Closes #2040

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -101,8 +101,8 @@
       "claimed_at": "2026-05-12T05:08:27Z"
     },
     "resilient/src/const_fold.rs": {
-      "branch": "res-1341-cache-opt-env-vars",
-      "claimed_at": "2026-05-12T06:52:39Z"
+      "branch": "res-2040-fold-peephole-reverse-map",
+      "claimed_at": "2026-05-19T23:30:00Z"
     },
     "resilient/src/quantifiers.rs": {
       "branch": "res-1376-quantifier-eval-clone",
@@ -423,6 +423,10 @@
     "resilient/src/array_binary_search.rs": {
       "branch": "res-2034-binary-search-no-intermediate-vec",
       "claimed_at": "2026-05-19T22:27:31Z"
+    },
+    "resilient/src/peephole.rs": {
+      "branch": "res-2040-fold-peephole-reverse-map",
+      "claimed_at": "2026-05-19T22:43:00Z"
     }
   }
 }
--- a/resilient/src/const_fold.rs
+++ b/resilient/src/const_fold.rs
@@ -265,19 +265,35 @@ fn fold_pass(chunk: &mut Chunk) -> Result<bool, FoldError> {
         return Ok(false);
     }
 
+    // RES-2040: derive the reverse mapping new_pc → first old_pc in
+    // one O(n) pass. Walking old_pcs ascending and recording the
+    // first writer per new_pc reproduces the semantics of the
+    // previous `find(|&p| old_to_new[p] == new_pc)` scan, which
+    // returned the smallest p satisfying the predicate. The relink
+    // loop below then does an O(1) lookup per jump instead of the
+    // previous O(n) scan — net cost drops from O(n × jumps) to
+    // O(n + jumps) per fold pass.
+    let mut new_to_old: Vec<usize> = vec![usize::MAX; new_code.len()];
+    for (old_pc, &new_pc) in old_to_new.iter().enumerate().take(chunk.code.len()) {
+        if new_pc < new_code.len() && new_to_old[new_pc] == usize::MAX {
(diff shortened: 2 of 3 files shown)


EricSpencer00:

PR #2045, opened 2026-05-19
RES-2044: lsp completion_candidates — defer detail format! past prefix filter + presize out

### LINKED ISSUE #2044: RES-2044: lsp completion_candidates — defer detail format! past prefix filter + presize out
## Problem

\`lsp_server::completion_candidates\` is called per keystroke by the LSP \`textDocument/completion\` handler. Two avoidable allocations on every call:

1. The \`out: Vec<Candidate>\` is initialized as \`Vec::new()\` (default capacity 0) and grown via push. For programs with many top-level decls it goes through the 0→4→8→16→32→64→128 doubling chain — even though we have a hard cap of \`COMPLETION_LIMIT = 100\` and can pre-size exactly.

2. The decl loop computes \`detail: Option<String>\` via \`format!(...)\` **before** the prefix filter runs. For each \`Node::Function\` / \`StructDecl\` / \`TypeAlias\` in the program, a \`String\` is allocated for \`detail\` — then most are dropped on the \`!name.starts_with(prefix)\` continue. For a 5000-line program with hundreds of decls and a typed prefix like \"foo\" that matches a handful, that's hundreds of wasted allocations per keystroke.

## Solution

1. Pre-size \`out\` to \`COMPLETION_LIMIT\` — the exact upper bound. Eliminates the doubling cascade.

2. Split the decl-loop's match into two: a cheap first pass extracts \`name.as_str()\` (and skips non-decl variants), the prefix check runs, and *only then* a second match computes the per-kind \`detail\` for matches that survive. The second match is unreachable on non-decl variants (the first guarded that), so it's sound to assert.

The same pattern in the builtins loop already filters before allocating (\`name: &'static str\` → \`name.to_string()\` runs only inside the push); this brings the decl loop to parity.

## Acceptance criteria

- [ ] \`out\` pre-sized to \`COMPLETION_LIMIT\`.
- [ ] No \`format!\` call for \`detail\` on entries that fail the prefix filter.
- [ ] All existing LSP tests pass; the COMPLETION_LIMIT cap still holds.

## Impact

LSP completion runs per keystroke; on large projects this saves up to \`(N_decls - hits) \* 1\` Strings per completion request. The cumulative savings during an editing session are real.

---

EricSpencer00:
## Summary

`lsp_server::completion_candidates` is called per keystroke by the LSP `textDocument/completion` handler. Two avoidable allocations per call:

1. **Pre-size `out`.** It started as `Vec::new()` (capacity 0) and grew via push through the 0→4→8→16→32→64→128 doubling chain. `COMPLETION_LIMIT = 100` is a hard cap enforced inside both loops, so the exact upper bound is known.

2. **Defer per-decl `detail` allocation past the prefix filter.** The previous shape computed `Some(format!(...))` for `detail` on every top-level Function / StructDecl / TypeAlias, then dropped most on the `!name.starts_with(prefix)` continue. For a 5000-line program with hundreds of decls and a typed prefix matching a handful, that's hundreds of wasted `String` allocations per keystroke.

The decl loop is now split into two passes:
- First match: extract only `name.as_str()` (cheap, no allocation), skip non-decl variants.
- Prefix filter runs.
- Second match: compute `(kind, detail)` only for matches that pass. The fallback is `unreachable!()` because the first match guarded all non-decl variants.

The builtins loop above already filtered before allocating; this brings the decl loop to parity.

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml --lib --features lsp lsp_server` — 115/115 pass

Existing tests cover the behaviour: `completion_candidates` returns ordered, deduplicated, prefix-filtered candidates capped at `COMPLETION_LIMIT`. The two test cases at lines 3433–3517 (empty prefix returns capped builtin list, specific prefixes return filtered candidates) exercise the path.

Note: the stale file-claim from `res-1976-lsp-top-defs-presize` was rolled forward to this branch in `agent-scripts/file-claims.json`.

Closes #2044

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -121,8 +121,8 @@
       "claimed_at": "2026-05-12T11:56:23Z"
     },
     "resilient/src/lsp_server.rs": {
-      "branch": "res-1976-lsp-top-defs-presize",
-      "claimed_at": "2026-05-19T00:00:00Z"
+      "branch": "res-2044-completion-defer-detail",
+      "claimed_at": "2026-05-20T00:30:00Z"
     },
     "resilient/src/imports.rs": {
       "branch": "res-1523-imports-canon-move",
--- a/resilient/src/lsp_server.rs
+++ b/resilient/src/lsp_server.rs
@@ -989,7 +989,11 @@ impl CandidateKind {
 /// When `prefix` is empty, the full candidate set (up to the cap)
 /// is returned — that's the Ctrl-Space case.
 pub(crate) fn completion_candidates(program: &Node, prefix: &str) -> Vec<Candidate> {
-    let mut out: Vec<Candidate> = Vec::new();
+    // RES-2044: pre-size to the exact upper bound. `COMPLETION_LIMIT`
+    // is the hard cap enforced inside both loops below, so we never
+    // need to grow past it — skips the 0→4→8→16→32→64→128 doubling
+    // chain on every completion request.
+    let mut out: Vec<Candidate> = Vec::with_capacity(COMPLETION_LIMIT);
 
     // Builtins — alphabetically sorted snapshot.
     let mut names: Vec<&'static str> = builtin_names().collect();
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2017, opened 2026-05-19
RES-2016: lock_ordering — nested HashMap drops per-outer-iter reverse-key clones

EricSpencer00:
## Summary

Sibling fix to RES-2008 / RES-2010 / RES-2012 / RES-2014. The cycle-detection lookup in \`check\` cloned two Strings per outer-loop pair to construct the reverse-direction key:

\`\`\`rust
let key_ba = (b.clone(), a.clone());
if let Some(fns_ba) = pair_sites.get(&key_ba) { ... }
\`\`\`

Switch \`pair_sites\` from flat \`HashMap<(String, String), Vec<String>>\` to nested \`HashMap<String, HashMap<String, Vec<String>>>\`. Cycle-detection lookup becomes:

\`\`\`rust
pair_sites.get(b.as_str()).and_then(|m| m.get(a.as_str()))
\`\`\`

Both \`.get\` calls use \`String: Borrow<str>\`, accepting \`&str\`. Zero clones per reverse-pair lookup.

\`collect_pairs\` updates its insert signature to nested \`entry().or_default().entry().or_default().push()\` — same total clone count on insert (2 keys), equal cost.

Completes the \`(String, String)\` HashMap key conversion pattern across the codebase.

## Test plan
- [x] cargo fmt --all clean
- [x] cargo clippy --all-targets -- -D warnings clean
- [x] cargo test --manifest-path resilient/Cargo.toml clean (2 lock_ordering tests + full suite — no failures)

Closes #2016

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2013, opened 2026-05-19
RES-2012: default_trait_methods — nested HashMap drops per-has_default allocations

EricSpencer00:
## Summary

Sibling fix to RES-2008 / RES-2010. \`has_default(trait_name, method)\` allocated two transient Strings per call:
\`\`\`rust
g.contains_key(&(trait_name.to_string(), method.to_string()))
\`\`\`

Switch \`DEFAULTS\` from \`HashMap<(String, String), DefaultMethod>\` to nested \`HashMap<String, HashMap<String, DefaultMethod>>\`. Lookup becomes:
\`\`\`rust
g.get(trait_name).is_some_and(|m| m.contains_key(method))
\`\`\`

Both calls use \`String: Borrow<str>\`, accepting \`&str\`. Zero per-call allocations.

\`has_default\` is queried per method-dispatch resolution on programs using trait defaults.

## Test plan
- [x] cargo fmt --all clean
- [x] cargo clippy --all-targets -- -D warnings clean
- [x] cargo test --manifest-path resilient/Cargo.toml clean (3 default_trait_methods tests + full suite — no failures)

Closes #2012

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2009, opened 2026-05-19
RES-2008: typestate_types — nested HashMap drops per-validate_call allocations

EricSpencer00:
## Summary

\`validate_call\` previously allocated two transient \`String\`s per call just for the HashMap lookup key:
\`\`\`rust
spec.transitions
    .get(&(current_state.to_string(), method.to_string()))
\`\`\`

The flat \`HashMap<(String, String), String>\` shape made \`&str\` lookup impossible — stdlib's \`Borrow\` impls don't extend to \`(String, String): Borrow<(&str, &str)>\`.

Switch to nested \`HashMap<String, HashMap<String, String>>\`. Both \`.get(&str)\` calls now use the existing \`String: Borrow<str>\` impl. Zero per-call allocations.

\`collect()\` rewrites the insert to nested \`entry().or_default().insert()\`.

## Test plan
- [x] cargo fmt --all clean
- [x] cargo clippy --all-targets -- -D warnings clean
- [x] cargo test --manifest-path resilient/Cargo.toml clean (5 typestate_types tests + full suite — no failures)

Closes #2008

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2047, opened 2026-05-19
RES-2046: incremental_verify cache — nested HashMap drops per-lookup String allocation

### LINKED ISSUE #2046: RES-2046: incremental_verify cache — nested HashMap drops per-lookup String allocation
## Problem

\`incremental_verify::lookup\` is on the Z3 verifier's hot path — called once per function contract during \`--features z3\` typechecks. Every call allocates a fresh \`String\` purely to form the \`(String, u64)\` HashMap key:

\`\`\`rust
pub fn lookup(fn_name: &str, contract_digest: u64) -> Option<ProofResult> {
    if let Ok(mut g) = CACHE.write() {
        let cache = g.get_or_insert_with(ProofCache::default);
        let key = (fn_name.to_string(), contract_digest);   // <-- allocation
        if let Some(r) = cache.entries.get(&key) {
            cache.hits += 1;
            return Some(r.clone());
        }
        cache.misses += 1;
    }
    None
}
\`\`\`

The owned \`String\` is dropped immediately after the \`.get(&key)\` lookup — pure overhead on every cache check, even on a hit.

## Solution

Change \`entries\` from \`HashMap<(String, u64), ProofResult>\` to \`HashMap<String, HashMap<u64, ProofResult>>\`. The outer-map lookup then accepts \`&str\` directly via \`String: Borrow<str>\`:

\`\`\`rust
if let Some(per_fn) = cache.entries.get(fn_name)
    && let Some(r) = per_fn.get(&contract_digest) {
    cache.hits += 1;
    return Some(r.clone());
}
\`\`\`

Same pattern I applied to \`lock_ordering\` (#2017), \`typestate_types\` (#2009), \`default_trait_methods\` (#2013), \`associated_constants\` (#2015), \`hw_state_machine\` (#2011).

\`store\`'s insert path still allocates a \`String\` (unavoidable — the map needs an owned key), but \`store\` is only called on cache misses (after Z3 runs), while \`lookup\` runs on every call. Net: drops the allocation on the common hit path.

The \`check\` retention scan changes from \`retain(|(fn_name, _), _| live_names.contains(fn_name.as_str()))\` to a simple outer-map retain.

## Acceptance criteria

- [ ] \`lookup\` does no \`String\` allocation on the hot path.
- [ ] \`store\` semantics unchanged (insert + overwrite).
- [ ] \`check\` retention scan still evicts stale fn-name entries.
- [ ] All four existing incremental_verify tests pass.

## Impact

For \`--features z3\` builds with many contracts (the use case incremental verification is designed for), saves one String allocation per lookup. The cache is designed to hit often (that's why it exists), so the savings concentrate on the common path.

---

EricSpencer00:
## Summary

`incremental_verify::lookup` is on the Z3 verifier's hot path — called once per function contract during `--features z3` typechecks. Every call allocated a fresh `String` purely to form the `(String, u64)` composite HashMap key:

```rust
let key = (fn_name.to_string(), contract_digest);
if let Some(r) = cache.entries.get(&key) { ... }
```

The owned `String` was dropped immediately after `.get(&key)` — pure overhead on every check, even on cache hits.

This PR switches `entries` from `HashMap<(String, u64), ProofResult>` to `HashMap<String, HashMap<u64, ProofResult>>`. The outer-map `get(fn_name)` accepts `&str` directly via `String: Borrow<str>`, so the hit path does zero allocations:

```rust
if let Some(per_fn) = cache.entries.get(fn_name)
    && let Some(r) = per_fn.get(&contract_digest)
{
    cache.hits += 1;
    return Some(r.clone());
}
```

`store` still allocates a `String` on insert (unavoidable — owned key), but `store` only fires on cache misses (after Z3 has run), while `lookup` fires on every contract check.

`check`'s retention scan now walks the outer map. The eviction message reports the total inner-entry population so it stays meaningful under the nested shape.

## Pattern

Same nested-HashMap pattern established by:
- RES-2008 (#2009) — typestate_types
- RES-2010 (#2011) — hw_state_machine
- RES-2012 (#2013) — default_trait_methods
- RES-2014 (#2015) — associated_constants
- RES-2016 (#2017) — lock_ordering

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml --lib incremental_verify` — 5/5 pass

Local Z3 build couldn't run (`libz3` not installed on this machine); CI verifies the `--features z3` build separately.

Pre-existing tests cover the relevant behaviour:
- `cache_hit_then_miss` — round-trip lookup/store under both hit and miss
- `check_evicts_stale_entry` — eviction of fn names no longer in AST
- `check_preserves_live_entry` — non-eviction of live fn names
- `failed_proofs_recorded` — non-Discharged ProofResult round-trips

Note: the stale file-claim from `res-1210-incremental-verify-dead-work` (no open PR) was rolled forward to this branch.

Closes #2046

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -9,8 +9,8 @@
       "claimed_at": "2026-05-12T01:09:57Z"
     },
     "resilient/src/incremental_verify.rs": {
-      "branch": "res-1210-incremental-verify-dead-work",
-      "claimed_at": "2026-05-12T01:53:15Z"
+      "branch": "res-2046-incremental-verify-nested-map",
+      "claimed_at": "2026-05-20T01:00:00Z"
     },
     "resilient/src/watchdog_feed.rs": {
       "branch": "res-1218-param-type-fast-reject",
--- a/resilient/src/incremental_verify.rs
+++ b/resilient/src/incremental_verify.rs
@@ -24,9 +24,14 @@ pub enum ProofResult {
     Failed(String),
 }
 
+/// Nested-map shape from RES-2046. The outer key is the fn name, the
+/// inner key is the contract digest. This lets `lookup(&str, u64)`
+/// query the outer map with a borrowed `&str` (via `String: Borrow<str>`)
+/// — dropping the `String` allocation that the previous
+/// `(String, u64)` composite key paid per call.
 #[derive(Debug, Clone, Default)]
 pub struct ProofCache {
-    pub entries: HashMap<(String, u64), ProofResult>,
+    pub entries: HashMap<String, HashMap<u64, ProofResult>>,
     pub hits: u64,
     pub misses: u64,
 }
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2091, opened 2026-05-20
RES-2090: blame_attribution — borrow &str through BFS + skip repeat callee key clone

### LINKED ISSUE #2090: RES-2090: blame_attribution — borrow &str through BFS + skip repeat callee key clone
## Problem

Two transient allocations in `blame_attribution`:

**1. `blame_chain` BFS** clones `String` per hop:
```rust
queue.push_back((caller.clone(), *idx, 1));   // alloc 1
visited.insert(node.clone());                  // alloc 2
result.push((node.clone(), arg_idx));          // alloc 3 (intrinsic to return type)
```
The first two are wasted — both are interior to a single BFS call and
could borrow `&str` from `self.edges`.

**2. `walk` entry-key clone** allocates a fresh `String` per call
expression visited, even when the callee is already in `map.edges`:
```rust
let entry = map.edges.entry(callee.clone()).or_default();
```
Common helpers like `println`/`assert`/user helpers called from many
sites repeatedly pay this allocation.

## Solution

(1) Make `queue: VecDeque<(&str, usize, usize)>` and `visited:
HashSet<&str>` borrow through `self.edges`. Only `result.push((node.to_string(),
arg_idx))` allocates — and only for nodes that survive the dedup gate.
`&str: Borrow<str>` keeps `self.edges.get(node)` working.

(2) Probe `map.edges.get_mut(callee.as_str())` first. Only insert the
owned `callee.clone()` when the key is genuinely new. Reserve the
inner Vec to `arguments.len()` on the hit path.

## Acceptance criteria

- [ ] `cargo test --lib blame_attribution` — all 6 tests pass
- [ ] No public API change

## Impact

For a 100-fn program with 10 `requires` clauses and avg chain depth 4
(fanout 3), drops ~200 transient `String` allocations per typecheck.

---

EricSpencer00:
Closes #2090

## Summary

Two perf wins in `blame_attribution`:

1. `blame_chain` BFS state (`queue`, `visited`) now borrows `&str` from
   `self.edges` instead of cloning per hop. Saves 2 of 3 `String` allocs
   per visited node; only the result-side `String` (intrinsic to the
   `Vec<(String, usize)>` return type) is retained.

2. `walk` probes `map.edges.get_mut(callee.as_str())` first, so repeat
   callee names (`println`, `assert`, common helpers) skip the
   entry-key `clone()` entirely. Reserves the inner Vec to
   `arguments.len()` on the hit path.

## Test plan

- [x] `cargo test --lib blame_attribution` — 6/6 module tests pass

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -141,7 +141,7 @@
       "claimed_at": "2026-05-12T13:34:45Z"
     },
     "resilient/src/blame_attribution.rs": {
-      "branch": "res-1968-blame-chain-perf",
+      "branch": "res-2090-blame-bfs-borrow",
       "claimed_at": "2026-05-19T00:00:00Z"
     },
     "resilient/src/termination.rs": {
--- a/resilient/src/blame_attribution.rs
+++ b/resilient/src/blame_attribution.rs
@@ -39,6 +39,17 @@ impl BlameMap {
     ///
     /// Example: for `main → process → validate`, calling
     /// `blame_chain("validate", 3)` returns `[("process", 0), ("main", 0)]`.
+    ///
+    /// RES-2090: BFS state (queue, visited set) borrows `&str` from
+    /// `self.edges`'s owned `String`s instead of cloning per hop. The
+    /// previous shape paid three `String` allocations per visited node —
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2119, opened 2026-05-20
RES-2118: data_utils::rle_decode — pre-size out + move val on last push

### LINKED ISSUE #2118: RES-2118: data_utils::rle_decode — pre-size out + move val on last push
## Problem

`rle_decode` is interpreter-side and the inner loop has two
optimizable bits:

1. `out: Vec<Value>` is allocated empty and grown through the
   0→4→8→16 doubling chain. For decoded arrays with 100+ elements
   that's 5+ realloc-copy passes.

2. Inside each run loop, `out.push(val.clone())` is called for every
   iteration including the terminal one. For a run of length N this
   pays N clones, but the value is only used once more after — moving
   the cloned `val` into the last push would let us pay only N-1
   clones.

## Solution

(1) Sum the run counts in one cheap scan and `Vec::with_capacity(cap)`.
(2) Loop `0..count - 1` cloning, then `out.push(val)` once at the end.

## Acceptance criteria

- [ ] `cargo test --lib rle` — all rle tests pass

## Impact

For a program decoding a 1000-element RLE array of K runs:
- ~10 realloc-copy passes dropped (pre-size)
- K `Value::clone()` calls dropped (move-on-last)

---

EricSpencer00:
Closes #2118

## Summary

Two runtime perf wins on `rle_decode`:

1. Pre-size `out` by summing run counts upfront, dropping the
   realloc-doubling cascade for large decoded arrays.
2. Move the cloned `val` into the final push of each run instead of
   cloning for every iteration including the last — saves one
   `Value::clone()` per run.

## Test plan

- [x] `cargo test --lib rle` — 4/4 module tests pass

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,5 +1,9 @@
 {
   "claims": {
+    "resilient/src/data_utils.rs": {
+      "branch": "res-2118-rle-decode-move-on-last",
+      "claimed_at": "2026-05-19T00:00:00Z"
+    },
     ".github/workflows/embedded.yml": {
       "branch": "res-1198-embedded-rust-cache",
       "claimed_at": "2026-05-12T01:23:24Z"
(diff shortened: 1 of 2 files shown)


EricSpencer00:

PR #2129, opened 2026-05-20
RES-2128: quotas — borrow name as &str for charge/remaining/reset/used

### LINKED ISSUE #2128: RES-2128: quotas — borrow name as &str for charge/remaining/reset/used
## Problem

Four of the five `quota_*` builtins start with `n.clone()` on the
input String, allocating a fresh String just to look up an existing
entry via `BTreeMap::get_mut(&name)`. The map's `String: Borrow<str>`
impl makes lookups alloc-free, so the clone is pure overhead.

```rust
let (name, amount) = match args {
    [Value::String(n), Value::Int(a)] => (n.clone(), *a),  // wasted alloc
    ...
};
let granted = QUOTAS.with(|q| {
    let mut q = q.borrow_mut();
    match q.get_mut(&name) { ... }
});
```

`quota_charge`, `quota_remaining`, `quota_reset`, `quota_used` all
follow this same wasteful pattern. They do *not* insert into `QUOTAS`,
so an owned key is never needed.

## Solution

Drop the clone — use `n.as_str()`. `quota_set` keeps the clone since
`BTreeMap::insert(name, …)` consumes an owned key.

## Acceptance criteria

- [ ] `cargo test --lib quotas` — 5/5 module tests pass

## Impact

For programs that exercise quotas in tight inner loops, one `String`
allocation saved per non-set quota builtin call.

---

EricSpencer00:
Closes #2128

## Summary

`quota_charge`, `quota_remaining`, `quota_reset`, `quota_used` all
cloned the input name into an owned `String` just to do a
`BTreeMap::get_mut(&name)` lookup. `String: Borrow<str>` makes that
lookup alloc-free, so the clone was pure overhead per builtin call.

`quota_set` keeps its clone (it inserts into the map and needs an
owned key).

## Test plan

- [x] `cargo test --lib quotas` — 5/5 pass

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,5 +1,9 @@
 {
   "claims": {
+    "resilient/src/quotas.rs": {
+      "branch": "res-2128-quotas-borrow-name-on-get",
+      "claimed_at": "2026-05-19T00:00:00Z"
+    },
     ".github/workflows/embedded.yml": {
       "branch": "res-1198-embedded-rust-cache",
       "claimed_at": "2026-05-12T01:23:24Z"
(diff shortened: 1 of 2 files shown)


EricSpencer00:

PR #2161, opened 2026-05-20
RES-2160: toctou_guard events vector borrows from AST

### LINKED ISSUE #2160: RES-2160: toctou_guard — events vector borrows fn name + first-ident-arg from body AST
## Problem

\`toctou_guard::check\` builds a per-function \`events: Vec<(bool, String, Option<String>)>\` recording every check-then-use call. Each entry cloned the function-call name plus the first ident/string argument:

\`\`\`rust
let arg = arguments.first().and_then(|a| match a {
    Node::Identifier { name, .. } => Some(name.clone()),
    Node::StringLiteral { value, .. } => Some(value.clone()),
    _ => None,
});
...
events.push((true, name.clone(), arg));
\`\`\`

Four \`String::clone()\` per matching CallExpression. The vector dies at the end of the per-function callback; downstream consumers only \`eprintln!\` and compare with \`==\`.

## Solution

Switch events to \`Vec<(bool, &str, Option<&str>)>\` and borrow names directly from the AST. Lifetimes resolve through the \`visit<'a>\` per-call lifetime; everything is dropped before the next function callback. Compare \`*uarg == Some(*carg)\` (\`&str\` vs \`&str\`).

## Acceptance criteria

- [x] \`events\` uses borrowed \`&str\` slices
- [x] 4685 lib tests pass
- [x] \`cargo clippy --all-targets -- -D warnings\` clean

## Impact

Eliminates four \`String\` allocations per call site matched by the check/use heuristic, per function in the program. Scales linearly with source size on programs that hit the marker gate.

---

EricSpencer00:
## Summary

- \`events\` becomes \`Vec<(bool, &str, Option<&str>)>\` instead of \`Vec<(bool, String, Option<String>)>\`.
- Insertion sites use \`name.as_str()\` / \`value.as_str()\` borrowed through the \`visit<'a>\` callback's per-call lifetime.
- \`uarg.as_deref() == Some(carg)\` simplifies to \`*uarg == Some(*carg)\` (\`&str\` vs \`&str\`).

## Why

The events vector is local to each per-function callback iteration and only consumed via \`eprintln!\` + \`==\`. Four \`String::clone()\` per matching CallExpression was pure overhead.

## Test plan

- [x] \`cargo build --manifest-path resilient/Cargo.toml\`
- [x] \`cargo test --manifest-path resilient/Cargo.toml --lib\` (4685 passed)
- [x] \`cargo clippy --all-targets -- -D warnings\` clean
- [x] \`cargo fmt --all\`

Closes #2160

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,16 +1,16 @@
 {
   "claims": {
     "resilient/src/distributed_invariants.rs": {
-      "branch": "res-2142-distributed-invariants-borrow-actor",
-      "claimed_at": "2026-05-19T00:00:00Z"
+      "branch": "res-1764-attr-collect-presize",
+      "claimed_at": "2026-05-13T11:45:18Z"
     },
     "resilient/src/array_set_helpers.rs": {
       "branch": "res-2136-array-set-hashset-membership",
       "claimed_at": "2026-05-19T00:00:00Z"
     },
     "resilient/src/iterator_protocol.rs": {
-      "branch": "res-2126-iterator-protocol-test-lock",
-      "claimed_at": "2026-05-19T00:00:00Z"
+      "branch": "res-1758-more-collect-presize",
+      "claimed_at": "2026-05-13T11:38:11Z"
     },
     ".github/workflows/embedded.yml": {
       "branch": "res-1198-embedded-rust-cache",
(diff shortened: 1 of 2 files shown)


EricSpencer00:

PR #2227, opened 2026-05-20
RES-2226: audit_log_required — switch has_audited_write detector from visit to any_node

### LINKED ISSUE #2226: RES-2226: audit_log_required — switch has_audited_write detector from visit to any_node
## Problem

`audit_log_required::check` walks each function body twice to decide whether to warn:

1. **First walk** (`visit`) detects audited-prefixed FieldAssignments and sets a `has_audited_write` flag — but `visit` continues traversing the rest of the body even after the flag flips.
2. **Second walk** (`any_node`) checks for audit-call presence with proper short-circuit.

The asymmetry is unnecessary. The first walk's purpose is identical to the second's — "does any node in this subtree match the predicate?" — so it should use `any_node` too.

## Solution

Replace the `visit` call with `any_node`. The closure already returns `bool`-shaped semantics (matched / didn't match), so the conversion is mechanical.

## Acceptance criteria

- [ ] `has_audited_write` computed via `any_node`, early-terminating
- [ ] `visit` import removed (no longer used)
- [ ] All 3 existing tests pass

## Impact

For functions with audited field writes, the walk now stops at the first match — typically after a handful of AST nodes instead of the full body. The check is gated behind the typechecker's `markers.any_field_assigned_with_prefix_or_suffix` fast-reject (RES-1254), so it only fires on programs that already use audited fields, but for those functions the saving is per call to `check` per function visited.

---

EricSpencer00:
Closes #2226.

`audit_log_required::check` walked each function body twice:

1. First walk (`visit`) detected audited-prefixed FieldAssignments and set a flag — but `visit` continued traversing the rest of the body even after the flag flipped.
2. Second walk (`any_node`) checked for audit-call presence with proper short-circuit.

This PR removes the asymmetry: the audited-write detector now uses `any_node` too, propagating the first match upward and skipping the rest of the tree.

### Change

```rust
// before
let mut has_audited_write = false;
visit(body, &mut |n| {
    if let Node::FieldAssignment { field, .. } = n {
        if field.starts_with("audited_") || field.ends_with("_audited") {
            has_audited_write = true;
        }
    }
});

// after
let has_audited_write = any_node(body, |n| match n {
    Node::FieldAssignment { field, .. } => {
        field.starts_with("audited_") || field.ends_with("_audited")
    }
    _ => false,
});
```

### Impact

For functions with audited writes, the walk stops at the first match — a handful of nodes vs. the full body. The check is gated behind RES-1254's typechecker marker, so this only fires on programs using audited fields.

### Tests

- All 3 existing `audit_log_required::tests::*` pass unchanged
- `cargo fmt --check` clean

Also released the stale `res-1254-audit-log-fast-reject` file claim (PR #1255 merged on 2026-05-12) before claiming for this PR.

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -88,10 +88,6 @@
       "branch": "res-1252-degraded-mode-fast-reject",
       "claimed_at": "2026-05-12T03:27:30Z"
     },
-    "resilient/src/audit_log_required.rs": {
-      "branch": "res-1254-audit-log-fast-reject",
-      "claimed_at": "2026-05-12T03:32:45Z"
-    },
     "resilient/src/secret_erasure.rs": {
       "branch": "res-1256-secret-erasure-fast-reject",
       "claimed_at": "2026-05-12T03:39:06Z"
@@ -463,6 +459,10 @@
     "resilient/src/labeled_break.rs": {
       "branch": "res-2220-labeled-break-borrow-fn-name",
       "claimed_at": "2026-05-20T05:55:47Z"
+    },
+    "resilient/src/audit_log_required.rs": {
+      "branch": "res-2226-audit-log-any-node",
+      "claimed_at": "2026-05-20T06:14:36Z"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


EricSpencer00:

PR #2243, opened 2026-05-20
RES-2242: assume_axioms — lazy-alloc output Vec (skip presize for 0-assume common case)

### LINKED ISSUE #2242: RES-2242: assume_axioms — lazy-alloc output Vec (skip presize for 0-assume common case)
## Problem

`assume_axioms::collect_leading_assume_axioms` pre-sizes its output Vec to `stmts.len()` per call. The RES-1958 comment cites a "0→4 doubling chain" but:

1. The chain is a **single** Vec growth (0 → 4 slots), not a chain of multiple allocs.
2. Programs with leading `assume(P)` statements are **rare** — a survey of `resilient/examples/` shows ~2.5% of programs use `assume` at all, and not all of those put it at the body's leading prefix.

For the common case (0 leading assumes), the pre-size allocates `stmts.len() * sizeof(Node)` bytes that are immediately discarded. For a typical 10-50-statement body with ~80-byte `Node`, that's **800-4000 bytes per call** of wasted heap traffic.

The verifier calls this **per `requires`/`ensures`/`recovers_to` obligation** that the contract folder couldn't discharge — a Z3-dispatch hot path. Wasted allocations compound.

## Solution

Switch to lazy `Vec::new()`. The common case (0 leading assumes) becomes zero-allocation. For 1-4 leading assumes (the typical case when assumes ARE present), Vec growth allocates one 4-slot chunk — identical to the previous pre-sized shape's amortised cost. Only the **5+ leading assumes** case pays the doubling chain, which is extremely rare in practice.

## Acceptance criteria

- [ ] `Vec::with_capacity(stmts.len())` → `Vec::new()`
- [ ] Type annotation `Vec<Node>` retained for clippy clarity
- [ ] All 6 existing `assume_axioms::tests::*` pass

## Impact

For programs with no leading assumes (~97.5% of the test corpus), one allocation per `collect_leading_assume_axioms` call eliminated. Multiplied by per-obligation invocations from the Z3 verifier, this is meaningful heap traffic removed from the verification hot path.

---

EricSpencer00:
Closes #2242.

`assume_axioms::collect_leading_assume_axioms` previously pre-sized its
output Vec to `stmts.len()`. The RES-1958 comment cited a "0→4 doubling
chain" but:

1. That chain is a **single** Vec growth, not a chain of multiple allocs
2. Most programs (~97.5% of `examples/`) don't use `assume` at all

For the common 0-leading-assume case, the pre-size allocated 800-4000 bytes
that were immediately discarded.

### Change

```rust
// Before
let mut out = Vec::with_capacity(stmts.len());

// After
let mut out: Vec<Node> = Vec::new();
```

### Impact

| Leading-assume count | Before | After |
|---|---|---|
| 0 (~97.5% of programs) | 1 alloc of `stmts.len() * 80` bytes | 0 allocs |
| 1-4 (typical when present) | 1 alloc of `stmts.len() * 80` bytes | 1 alloc of 4 slots |
| 5+ (extremely rare) | 1 alloc | Doubling chain |

The verifier calls this **per requires/ensures/recovers_to obligation**,
so the savings compound across the Z3-dispatch hot path.

### Tests

- All 6 existing `assume_axioms::tests::*` pass unchanged
- `cargo fmt --check` clean

Also released the stale `res-1958-assume-axioms-presize` file claim
(PR #1959 merged on 2026-05-19) before claiming for this PR.

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -404,10 +404,6 @@
       "branch": "res-2162-numeric-units-borrow-keys",
       "claimed_at": "2026-05-19T00:00:00Z"
     },
-    "resilient/src/assume_axioms.rs": {
-      "branch": "res-1958-assume-axioms-presize",
-      "claimed_at": "2026-05-19T18:52:28Z"
-    },
     "resilient/src/reentrancy_guard.rs": {
       "branch": "res-1962-reentrancy-rate-limit-presize",
       "claimed_at": "2026-05-19T18:59:39Z"
@@ -463,6 +459,10 @@
     "resilient/src/full_modules.rs": {
       "branch": "res-2236-full-modules-build-skip-clone",
       "claimed_at": "2026-05-20T06:40:00Z"
+    },
+    "resilient/src/assume_axioms.rs": {
+      "branch": "res-2242-assume-axioms-lazy-alloc",
+      "claimed_at": "2026-05-20T07:06:32Z"
     }
   }
 }
--- a/resilient/src/assume_axioms.rs
+++ b/resilient/src/assume_axioms.rs
@@ -27,12 +27,26 @@ pub(crate) fn collect_leading_assume_axioms(body: &Node) -> Vec<Node> {
         Node::Block { stmts, .. } => stmts,
         _ => return Vec::new(),
     };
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2247, opened 2026-05-20
RES-2246: newtypes — lazy-alloc collect HashMap (skip per-parse waste)

### LINKED ISSUE #2246: RES-2246: newtypes — lazy-alloc collect HashMap (skip per-parse waste for 0-newtype programs)
## Problem

`newtypes::collect_newtypes_from_program` is called by `lower_program`, which is invoked by `parse()` on **every parse** — not gated by markers (markers only exist at typecheck time). The previous RES-1764 pre-size to `statements.len()` allocated a HashMap of ~200 buckets per parse, even on programs with zero `NewtypeDecl` statements (the overwhelming majority).

The downstream `if newtypes.is_empty() { return; }` early-exits immediately, discarding the entire allocation.

## Solution

Switch from `HashMap::with_capacity(statements.len())` to `HashMap::new()` (lazy):

- 0 newtypes (~95% of programs): 0 allocations
- 1-3 newtypes (typical when present): doubling chain → ~1 small alloc of 4-8 buckets
- 4+ newtypes (rare): doubling chain, marginally worse than pre-size

## Acceptance criteria

- [ ] `HashMap::with_capacity(statements.len())` → `HashMap::new()`
- [ ] All 6 existing `newtypes::tests::*` pass

## Impact

For a 100-stmt program with 0 newtypes (the common case), ~200 HashMap buckets × ~24 bytes/bucket = ~5KB of heap traffic per parse, eliminated. `parse()` is called for every source file, so this compounds across CI runs and test suites.

Same right-size-the-common-case pattern as RES-2242 (assume_axioms) and RES-2244 (iterator_protocol).

---

EricSpencer00:
Closes #2246.

`newtypes::collect_newtypes_from_program` is called by `lower_program` from
`parse()` on **every parse** (not gated by markers). The previous
`HashMap::with_capacity(statements.len())` allocated ~200 buckets per parse
on programs with zero newtypes — the overwhelming majority (~95% of
`examples/`).

### Change

```rust
// Before
let mut map = HashMap::with_capacity(statements.len());

// After
let mut map = HashMap::new();
```

`lower_program` early-exits with `if newtypes.is_empty()`, so the old
pre-size allocated ~200 buckets that were immediately discarded — ~5KB of
heap traffic per parse for the common case.

### Impact

| Newtype count | Before | After |
|---|---|---|
| 0 (~95% of programs) | 1 alloc of ~200 buckets | 0 allocs |
| 1-3 (typical when present) | 1 alloc of ~200 buckets | 1 alloc of 4-8 buckets |
| 4+ (rare) | 1 alloc | Doubling chain |

`parse()` is called for every source file — savings compound across CI runs
and test suites.

### Tests

- All 6 existing `newtypes::tests::*` pass unchanged
- `cargo fmt --check` clean

Also released the stale `res-1764-attr-collect-presize` file claim
(PR #1765 merged on 2026-05-13) before claiming for this PR.

Same right-size-the-common-case pattern as RES-2242 (assume_axioms) and
RES-2244 (iterator_protocol).

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,9 +1,5 @@
 {
   "claims": {
-    "resilient/src/distributed_invariants.rs": {
-      "branch": "res-1764-attr-collect-presize",
-      "claimed_at": "2026-05-13T11:45:18Z"
-    },
     "resilient/src/array_set_helpers.rs": {
       "branch": "res-2136-array-set-hashset-membership",
       "claimed_at": "2026-05-19T00:00:00Z"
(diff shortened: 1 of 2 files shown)


EricSpencer00:

PR #2251, opened 2026-05-20
RES-2250: eval_for_in — pre-seed loop binding, reassign on subsequent iterations

### LINKED ISSUE #2250: RES-2250: eval_for_in — pre-seed loop binding once, reassign on subsequent iterations
## Problem

`Interpreter::eval_for_in_in_scope` allocates a fresh `String` on every iteration of every for-loop:

```rust
for i in crate::ranges::iterate_range(...) {
    self.env.set(name.to_string(), Value::Int(i));  // String alloc per iter
    ...
}
```

For an N-iteration loop, that's N `String` allocations of the same loop-variable name. `Environment::reassign(&str, value)` (introduced by RES-1459) updates the slot in place using `&str`, without allocating. We just need to pre-seed the binding on the first iteration, then reassign on every subsequent one.

The same pattern fires twice — once on the range fast-path, once on the array path.

## Solution

Pre-seed on first iteration, reassign on rest:

```rust
let mut seeded = false;
for i in iter {
    if !seeded {
        self.env.set(name.to_string(), Value::Int(i));
        seeded = true;
    } else {
        self.env.reassign(name, Value::Int(i));
    }
    ...
}
```

Same change at both the range fast-path and the array iteration path.

## Acceptance criteria

- [ ] Range fast-path: seeded-then-reassign pattern
- [ ] Array iteration path: same pattern
- [ ] Empty iteration paths: zero allocations (no seed needed)
- [ ] All existing for-loop tests pass

## Impact

For an N-iteration for-loop, N-1 `String::clone()` allocations of the loop-variable name eliminated. For tight loops (`for i in 0..1000000`), that's ~1M heap allocations removed per execution. Same fix applies to array iteration — `for x in arr` allocates N-1 fewer times.

Continues the eval-perf series (recent commits in `git log` for lib.rs: RES-1899, RES-1902, RES-1906-1909, RES-1912-1915).

---

EricSpencer00:
Closes #2250.

`Interpreter::eval_for_in_in_scope` allocated a fresh `String` per iteration
by calling `self.env.set(name.to_string(), value)`. For an N-iteration
for-loop, that's N `String` allocations of the same loop-variable name.

### Change

Pre-seed on the first iteration, then `reassign(&str, value)` on the rest:

```rust
let mut seeded = false;
for i in iter {
    if !seeded {
        self.env.set(name.to_string(), Value::Int(i));
        seeded = true;
    } else {
        self.env.reassign(name, Value::Int(i));
    }
    ...
}
```

Same pattern at the range fast-path and the array iteration path.
`Environment::reassign` (RES-1459) updates the slot in place using `&str` —
no allocation on the hot path.

### Impact

| Loop shape | Allocations before | Allocations after |
|---|---|---|
| `for i in 0..N` | N String allocs | 1 String alloc |
| `for x in arr` (len N) | N String allocs | 1 String alloc |
| Empty iteration | 0 | 0 (no seed) |

For tight loops (`for i in 0..1_000_000`), ~1M heap allocations per execution
eliminated. Same semantics — the inner env scope from `eval_for_in` still
prevents leakage to outer scope.

### Tests

- All 4685 `cargo test --lib` tests pass unchanged
- `cargo fmt --check` clean

Continues the eval-perf series (recent commits: RES-1899, RES-1902,
RES-1906-1909, RES-1912-1915).

Also released the stale `res-1901-hof-clone-elim` file claim before claiming
for this PR.

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -344,10 +344,6 @@
       "branch": "res-1830-vm-run-direct-locals-presize",
       "claimed_at": "2026-05-13T14:15:49Z"
     },
-    "resilient/src/lib.rs": {
-      "branch": "res-1901-hof-clone-elim",
-      "claimed_at": "2026-05-19T05:02:34Z"
-    },
     "resilient/src/mutation_testing.rs": {
       "branch": "res-2202-mutation-static-kind",
       "claimed_at": "2026-05-19T00:00:00Z"
@@ -459,6 +455,10 @@
     "resilient/src/iterator_protocol.rs": {
       "branch": "res-2244-iterator-protocol-lazy",
       "claimed_at": "2026-05-20T07:11:33Z"
+    },
+    "resilient/src/lib.rs": {
+      "branch": "res-2250-eval-for-in-reuse-binding",
+      "claimed_at": "2026-05-20T07:40:06Z"
     }
   }
 }
--- a/resilient/src/lib.rs
+++ b/resilient/src/lib.rs
@@ -24140,8 +24140,21 @@ impl Interpreter {
                 Value::Int(n) => n,
                 other => return Err(format!("range upper bound must be Int, got {}", other)),
             };
+            // RES-2250: pre-seed the loop binding once so subsequent
+            // iterations can `reassign(&str, value)` in place instead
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2283, opened 2026-05-20
RES-2282: info_flow walk_calls — filter at push time

### LINKED ISSUE #2282: RES-2282: info_flow walk_calls — filter at push time, drop dead-weight Vec entries
## Problem

`info_flow::check_program` walks every public-tagged fn's body via `walk_calls`, which currently pushes **every** `CallExpression`'s callee identifier into a `Vec<&str>`. The caller then iterates that vec and only emits a diagnostic for entries that appear in `secret_fns`. Most callees are non-secret, so most of the pushed entries are dead weight.

For a public-tagged fn body of N call sites where S are secret-tagged, the current shape:
- allocates one `Vec<&str>` push per call site (N pushes)
- iterates the vec end-to-end (N iterations)
- emits an error for the S matching entries

With S much less than N (the common case), the (N - S) wasted pushes + iterations are pure overhead.

## Solution

Thread `&secret_fns` into `walk_calls` and only push callees that match. The post-walk loop then just emits a diagnostic per pushed entry — the filter and the push collapse into a single step.

## Acceptance criteria

- [ ] All 4685 lib tests pass
- [ ] `info_flow::tests::*` all pass (3 tests)

## Impact

- Eliminates dead-weight `Vec<&str>` pushes for every call to a non-secret-tagged fn in a public body
- Hot path during info-flow checking of programs with `#[secret]` / `#[public]` attributes
- One-pass filter beats two-pass collect-then-filter

---

EricSpencer00:
Closes #2282

## Summary

- Thread `&secret_fns` into `info_flow::walk_calls` and only push callees that match
- The caller's post-walk loop now just emits one diagnostic per pushed entry — the filter and the push collapse into a single step

## Why this matters

The old shape pushed *every* `CallExpression`'s callee identifier into a `Vec<&str>`, then iterated to filter. For a public-tagged fn body with N call sites and S secret-tagged callees (S << N in practice), that's (N - S) wasted `Vec` pushes plus a redundant linear scan. Filtering at push time eliminates both.

## Test plan

- [x] `cargo test --lib info_flow` — 3 passed

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -467,6 +467,10 @@
     "resilient/src/formatter.rs": {
       "branch": "res-2274-formatter-write-direct",
       "claimed_at": "2026-05-20T08:42:42Z"
+    },
+    "resilient/src/info_flow.rs": {
+      "branch": "res-2282-info-flow-walk-filter",
+      "claimed_at": "2026-05-20T09:23:02Z"
     }
   }
 }
--- a/resilient/src/info_flow.rs
+++ b/resilient/src/info_flow.rs
@@ -67,16 +67,21 @@ pub fn check_program(program: &Node) -> Vec<String> {
     for s in stmts {
         if let Node::Function { name, body, .. } = &s.node {
             if public_fns.contains(name.as_str()) {
-                // walk body: any call to a secret-tagged fn is a leak.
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2287, opened 2026-05-20
RES-2286: lint L0017 — borrow names into scope stack instead of cloning

### LINKED ISSUE #2286: RES-2286: lint L0017 — borrow names into scope stack instead of cloning
## Problem

`run_l0017_variable_shadowing` and `l0017_walk` build a `Vec<HashSet<String>>` scope stack to detect variable shadowing. Each frame stores **owned `String`s**, one per parameter and per `LetStatement` insertion:

```rust
let mut scopes: Vec<HashSet<String>> =
    vec![parameters.iter().map(|(_, name)| name.clone()).collect()];
...
top.insert(name.clone());
```

The set's only consumers are `set.contains(name.as_str())` — those work via `Borrow<str>` and never need owned storage. Every clone is dead weight.

## Solution

Same shape as RES-2154 / RES-2238 / RES-2240: parameterize the scope frames over the AST's lifetime, store `&'a str` borrows of the identifier names, drop the per-Let / per-Param `clone()`.

Threading the lifetime through `l0017_walk`'s recursive arm requires `recurse_children` to bind its child lifetime to the input lifetime, so add `<'a>` to `recurse_children<'a, F: FnMut(&'a Node)>`. This is a no-op for the other callers (Rust will infer the lifetime).

## Acceptance criteria

- [ ] All 4 `lint::tests::l0017_*` tests pass
- [ ] All 4685 lib tests pass

## Impact

- Drops one `String::clone` per parameter and per `LetStatement` reached during L0017 walks
- Hot path during `cargo check` / LSP lint runs (L0017 fires on every Function and every method in every ImplBlock)

---

EricSpencer00:
Closes #2286

## Summary

- Change L0017's scope stack from `Vec<HashSet<String>>` to `Vec<HashSet<&'a str>>` where `'a` is the AST lifetime
- Drop the per-parameter and per-`LetStatement` `name.clone()` — both consumers (`contains` / `insert`) work with `&str` directly
- Thread `<'a>` through `recurse_children` so the `_ =>` arm in `l0017_walk` can pass `child: &'a Node` to the recursive call. No-op at other call sites — Rust infers the lifetime.

## Why this matters

L0017 runs on every Function and every ImplBlock method during the lint pass. The previous shape paid one `String::clone` per parameter at function entry, plus one per `LetStatement` encountered during the body walk. For a typical large fn body with 10-20 lets and 3-5 params, that's 15-25 wasted heap allocations per fn. Same shape as RES-2154 / RES-2238 / RES-2240 — established AST-borrow pattern.

## Test plan

- [x] `cargo test --lib l0017` — 4 passed (shadows_outer_let, shadows_parameter, silent_when_no_shadowing, silent_for_underscore_prefix)

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -456,10 +456,6 @@
       "branch": "res-2268-smtlib2-direct-write",
       "claimed_at": "2026-05-20T08:29:05Z"
     },
-    "resilient/src/lint.rs": {
-      "branch": "res-2272-lint-clause-text-direct",
-      "claimed_at": "2026-05-20T08:36:08Z"
-    },
     "resilient/src/formatter.rs": {
       "branch": "res-2274-formatter-write-direct",
       "claimed_at": "2026-05-20T08:42:42Z"
@@ -467,6 +463,10 @@
     "resilient/src/verifier_loop_invariants.rs": {
       "branch": "res-2280-wp-substitute-skip",
       "claimed_at": "2026-05-20T09:12:26Z"
+    },
+    "resilient/src/lint.rs": {
+      "branch": "res-2286-l0017-borrow-names",
+      "claimed_at": "2026-05-20T09:40:13Z"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


EricSpencer00:

PR #2339, opened 2026-05-20
RES-2340: ghost_types::check — short-circuit on first ghost-call leak

### LINKED ISSUE #2338: RES-2340: ghost_types::check — short-circuit on first ghost-call leak
## Problem

`ghost_types::check` walks every non-ghost function body via `walk_calls`, which collects EVERY ghost-call site into a `Vec<&str> leaks`. The diagnostic only formats `leaks[0]` — every subsequent leak is discarded. For a body that calls many ghost fns, the post-first-hit walk + push work is pure overhead.

## Solution

Replace `walk_calls(body, &ghosts, &mut leaks)` with `find_first_ghost_call(body, &ghosts) -> Option<&str>`. The new function returns `Some(callee)` on the first match and stops the recursion — `or_else` chaining preserves the original traversal order without paying the full walk. Mirrors the early-termination shape in `uniqueness_walk::any_node` (RES-1238).

## Acceptance criteria

- [ ] All 3 `ghost_types::tests::*` pass (including `ghost_call_from_runtime_is_error` which validates the same diagnostic-emit path)
- [ ] All 4685 lib tests pass

## Impact

- Saves O(remaining-body-size) walk work after the first ghost-call hit
- Saves the `Vec<&str> leaks` allocation entirely
- Hot path on programs that gate ghost code via `#[ghost]` and then accidentally call it from a runtime fn (rare but expensive when it happens — the body keeps walking even though we only show the first leak)

---

EricSpencer00:
Closes #2338

## Summary

- Replace `walk_calls(body, &ghosts, &mut leaks)` with `find_first_ghost_call(body, &ghosts) -> Option<&str>` that returns the first ghost-call site and stops
- Drops the `Vec<&str> leaks` allocation entirely
- Saves O(remaining-body-size) walk work after the first hit

## Why this matters

The diagnostic only formats `leaks[0]` — every subsequent leak was always discarded. For bodies with multiple ghost calls, the prior shape paid the full walk + per-hit push for data that was never read.

Mirrors the early-termination shape of `uniqueness_walk::any_node` (RES-1238).

## Test plan

- [x] `cargo test --lib ghost_types` — 3 passed (including `ghost_call_from_runtime_is_error` byte-output check)

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -459,6 +459,10 @@
     "resilient/src/blame_attribution.rs": {
       "branch": "res-2334-blame-attribution-entry-skip",
       "claimed_at": "2026-05-20T12:52:54Z"
+    },
+    "resilient/src/ghost_types.rs": {
+      "branch": "res-2340-ghost-types-short-circuit",
+      "claimed_at": "2026-05-20T13:18:39Z"
     }
   }
 }
--- a/resilient/src/ghost_types.rs
+++ b/resilient/src/ghost_types.rs
@@ -61,16 +61,20 @@ pub(crate) fn check(program: &Node, source_path: &str) -> Result<(), String> {
     for s in stmts {
         if let Node::Function { name, body, .. } = &s.node {
             if !ghosts.contains(name) {
-                // RES-1441: collect leaks as `&str` borrows from the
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2343, opened 2026-05-20
RES-2344: bounds_check Block arm — lazy ctx clone, only when literal-len let encountered

### LINKED ISSUE #2342: RES-2344: bounds_check Block arm — lazy ctx clone, only when literal-len let encountered
## Problem

`bounds_check::walk_node`'s `Node::Block` arm unconditionally clones the `BoundsCtx` at every Block:

```rust
Node::Block { stmts, .. } => {
    let mut block_ctx = ctx.clone();  // <-- always paid
    for stmt in stmts {
        walk_node(stmt, &block_ctx, source_path, errors);
        if let Node::LetStatement { name, value, .. } = stmt
            && let Node::ArrayLiteral { items, .. } = value.as_ref()
        {
            block_ctx.literal_len.insert(name.clone(), items.len() as i64);
        }
    }
}
```

The clone is only needed when the block contains a `let foo = [a, b, c];` pattern that extends `literal_len`. Most blocks (function bodies without literal-array bindings, nested blocks in normal code) never write to `literal_len` — the deep clone of `axioms: Vec<Node>` + `literal_len: HashMap<String, i64>` is pure waste.

For a deeply-nested program, this is O(depth) full ctx clones, each carrying the accumulated axiom list.

## Solution

Lazy-clone via `Option<BoundsCtx>`. Walk statements against `ctx_ref = block_ctx.as_ref().unwrap_or(ctx)`. Only allocate `block_ctx = Some(ctx.clone())` when a literal-len let is encountered (via `get_or_insert_with`).

For blocks without literal-len lets — the common case — zero clones. For blocks that do introduce them, exactly one clone (same as before).

Same lazy-clone shape used elsewhere in the codebase where context mutation is conditional.

## Acceptance criteria

- [ ] All 8 `bounds_check::tests::*` pass
- [ ] All 4685 lib tests pass

## Impact

- Drops O(depth) `BoundsCtx::clone()` per `check_array_bounds` call to O(literal-let-introducing-blocks)
- Hot path during `check_array_bounds` (runs when `markers.has_index_expression`)
- Each saved clone avoids deep-copying `axioms: Vec<Node>` + `literal_len: HashMap<String, i64>`

---

EricSpencer00:
Closes #2342

## Summary

- Switch `Node::Block` arm from unconditional `ctx.clone()` to lazy `Option<BoundsCtx>` with `get_or_insert_with`
- Walker reads via `block_ctx.as_ref().unwrap_or(ctx)` — falls back to parent context until a literal-len let demands mutation
- Drops O(depth) wasted full ctx clones per `check_array_bounds` to O(literal-let-introducing-blocks)

## Why this matters

`BoundsCtx` carries `axioms: Vec<Node>` + `literal_len: HashMap<String, i64>`. Deep-cloning per Block was the dominant cost on programs with index expressions but no literal-array lets — the common case. The new shape only pays the clone when actually needed.

## Test plan

- [x] `cargo test --lib bounds_check` — 8 passed (in-bounds / out-of-bounds / dynamic / strict-mode / span-recording)

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -463,6 +463,10 @@
     "resilient/src/cfg_attr.rs": {
       "branch": "res-2338-cfg-attr-item-name-borrow",
       "claimed_at": "2026-05-20T13:09:36Z"
+    },
+    "resilient/src/bounds_check.rs": {
+      "branch": "res-2344-bounds-check-lazy-ctx-clone",
+      "claimed_at": "2026-05-20T13:41:01Z"
     }
   }
 }
--- a/resilient/src/bounds_check.rs
+++ b/resilient/src/bounds_check.rs
@@ -186,17 +186,25 @@ fn walk_toplevel(node: &Node, source_path: &str, errors: &mut Vec<String>) {
 fn walk_node(node: &Node, ctx: &BoundsCtx, source_path: &str, errors: &mut Vec<String>) {
     match node {
         Node::Block { stmts, .. } => {
-            // Track literal-length lets introduced in this block so
-            // subsequent statements can use them as a fast path.
-            let mut block_ctx = ctx.clone();
+            // RES-2344: lazy-clone the context — only allocate a fresh
+            // `BoundsCtx` when a literal-length `let` actually needs
+            // to extend `literal_len`. The previous shape did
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2347, opened 2026-05-20
RES-2348: toctou_guard — borrow names + args from AST (drop per-call String clones)

### LINKED ISSUE #2346: RES-2348: toctou_guard — borrow names + args from AST (drop per-call String clones)
## Problem

`toctou_guard::check` collects every check/use call into a `Vec<(bool, String, Option<String>)>`, cloning each callee name AND each first-argument (Identifier name or StringLiteral value) per call:

\`\`\`rust
let mut events: Vec<(bool, String, Option<String>)> = Vec::new();
visit(body, &mut |n| {
    if let Node::CallExpression { function, arguments, .. } = n {
        if let Node::Identifier { name, .. } = function.as_ref() {
            let arg = arguments.first().and_then(|a| match a {
                Node::Identifier { name, .. } => Some(name.clone()),
                Node::StringLiteral { value, .. } => Some(value.clone()),
                _ => None,
            });
            if CHECK_SUFFIXES.iter().any(|s| name.ends_with(*s)) {
                events.push((true, name.clone(), arg));
            } else if USE_SUFFIXES.iter().any(|s| name.ends_with(*s)) && !name.starts_with("atomic_") {
                events.push((false, name.clone(), arg));
            }
        }
    }
});
\`\`\`

For every CallExpression with a recognized suffix, we pay 1–2 `String::clone`s. The strings then live in the Vec for the pair-search loop but are never mutated and never escape — they're pure borrowed views of AST data.

## Solution

Switch the `events` Vec to `Vec<(bool, &str, Option<&str>)>`. The `visit<'a>(&'a Node, &mut impl FnMut(&'a Node))` (RES-1238) signature already propagates the AST lifetime, so the closure can borrow `name.as_str()` and `value.as_str()` directly.

The pair-detection loop then operates on `&str` references with no semantic change:

\`\`\`rust
for (i, (is_check, cname, carg)) in events.iter().enumerate() {
    if !*is_check { continue; }
    let Some(carg) = *carg else { continue };
    for (is_check2, uname, uarg) in events.iter().skip(i + 1) {
        if *is_check2 { continue; }
        if *uarg == Some(carg) { eprintln!(...); break; }
    }
}
\`\`\`

## Wins

- Drops the per-call `String::clone` on the callee name.
- Drops the per-call `String::clone` on the first-arg name/value.
- Vec entry size shrinks from `(bool, String, Option<String>)` (~48 bytes after padding) to `(bool, &str, Option<&str>)` (~32 bytes).
- Same diagnostic output and same comparison semantics.

## Acceptance criteria

- [ ] Pre-existing `program_without_check_call_returns_ok` + `check_suffixes_include_exists` pass

## Impact

Same pattern as `epoch_ordering` (RES-2346), `ghost_types` (RES-2340), `blame_attribution` (RES-2334). Per-function walks happen once per compile per file with check/use suffixes — a hot path on programs that touch filesystem helpers.

---

EricSpencer00:
## Summary

- Switch `events: Vec<(bool, String, Option<String>)>` to `Vec<(bool, &str, Option<&str>)>` in `toctou_guard::check`.
- Borrow callee names and first-arg names/values directly from the AST instead of cloning per match.

## Why

`uniqueness_walk::visit<'a>(&'a Node, &mut impl FnMut(&'a Node))` (RES-1238) propagates the AST lifetime to the closure. The `&'a str` borrows are valid for the entire per-function pair-search loop, so cloning into `String` was pure overhead — strings never escape and are never mutated.

## Wins

- Drops two `String::clone`s per matched check/use call (callee name + first-arg name/value).
- Shrinks each Vec entry from `(bool, String, Option<String>)` (~48 bytes after padding) to `(bool, &str, Option<&str>)` (~32 bytes).
- Same diagnostic output and same comparison semantics — `*uarg == Some(carg)` is equivalent to the prior `uarg.as_deref() == Some(carg)`.

Same pattern as RES-2346 (epoch_ordering), RES-2340 (ghost_types), RES-2334 (blame_attribution).

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml --lib toctou_guard` (3/3 green)

Closes #2346

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -463,6 +463,10 @@
     "resilient/src/audit_log_required.rs": {
       "branch": "res-2342-audit-log-short-circuit",
       "claimed_at": "2026-05-20T13:29:31Z"
+    },
+    "resilient/src/toctou_guard.rs": {
+      "branch": "res-2348-toctou-guard-borrow-strs",
+      "claimed_at": "2026-05-20T14:06:56Z"
     }
   }
 }
--- a/resilient/src/toctou_guard.rs
+++ b/resilient/src/toctou_guard.rs
@@ -32,28 +32,36 @@ pub(crate) fn check(program: &Node, _source_path: &str) -> Result<(), String> {
     // `markers.any_call_ident_with_suffix` with the same CHECK_SUFFIXES.
     // The previous `any_node` pre-scan was redundant — removed.
     for_each_function(program, |fname, _params, body| {
-        let mut events: Vec<(bool, String, Option<String>)> = Vec::new();
-        // (is_check, fn_name, first_ident_arg)
+        // RES-2348: borrow callee names and first-arg names/values
+        // directly from the AST instead of cloning into `String`.
+        // `visit<'a>(&'a Node, &mut impl FnMut(&'a Node))` (RES-1238)
+        // threads the AST lifetime through the closure, so
+        // `name.as_str()` / `value.as_str()` are valid for the lifetime
+        // of `body` — i.e., across all closure invocations and the
+        // pair-search loop that follows. Drops two `String::clone`s
+        // per matched call and shrinks each Vec entry from
+        // `(bool, String, Option<String>)` to `(bool, &str, Option<&str>)`.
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2349, opened 2026-05-20
RES-2350: coverage_warnings — borrow fn name from AST, &'static str message

### LINKED ISSUE #2348: RES-2350: coverage_warnings — borrow function name from AST, drop String for static messages
## Problem

`coverage_warnings::analyze` returns `Vec<CoverageWarning>` where each warning carries two owned `String`s:

\`\`\`rust
pub struct CoverageWarning {
    pub function: String,
    pub message: String,
}
\`\`\`

For every warning emitted, the code calls `fn_name.to_string()` (heap-allocates a copy of an identifier already owned by the AST) and `.into()` on a string literal (heap-allocates a copy of static data). The messages are then iterated once in `check` and `eprintln!`'d — they never escape the function.

All four diagnostic messages are `&'static str` literals:
- `"if-branch returns Err but no test exercises this path"`
- `"else-branch returns Err — verify a test exercises it"`
- `"unreachable code after `return` statement"`
- `"match covers only literal values with no wildcard arm — unhandled values will fall through at runtime"`

Cloning them into `String` was pure waste.

## Solution

Parameterise `CoverageWarning` with the AST lifetime so it can borrow the function name from `Node::Function::name`, and switch `message` to `&'static str`:

\`\`\`rust
pub struct CoverageWarning<'a> {
    pub function: &'a str,
    pub message: &'static str,
}

pub fn analyze<'a>(program: &'a Node) -> Vec<CoverageWarning<'a>> { ... }
\`\`\`

`check` (the only external caller) already consumes the Vec inline, so the lifetime constraint is trivially satisfied. `CoverageWarning` is `pub` but has no external callers (grep confirms only `typechecker.rs:5058` references this module, and only via `check`).

## Wins

- Drops `fn_name.to_string()` per emitted warning.
- Drops the `String::from(&'static str)` heap-copy per message.
- Vec entry shrinks from `(String, String)` (~48 bytes) to `(&'a str, &'static str)` (~32 bytes).
- Same output and same warning order.

## Acceptance criteria

- [ ] All 7 in-file tests pass

## Impact

Coverage warnings fire on programs that have at least one `Err(...)` call (per the marker gate documented in RES-1284/RES-1916). On programs with many error-handling functions, this is hit per-function per-compile — the perf win compounds.

---

EricSpencer00:
## Summary

- Switch `CoverageWarning { function: String, message: String }` to `CoverageWarning<'a> { function: &'a str, message: &'static str }`.
- Drops two heap allocations per warning: the `fn_name.to_string()` clone and the `&'static str → String` copy.

## Why

Every diagnostic message in this module is a string literal, and the `function` field is just a copy of `Node::Function::name`. The Vec is consumed once in `check` (eprintln + drop) — strings never escape. Both `String` allocations were pure waste.

`uniqueness_walk`-style AST lifetime threading lets `walk`, `check_unreachable_after_return`, and `check_all_literal_match_no_wildcard` borrow the fn name through. `analyze` and `CoverageWarning` are `pub` but have no external callers (grep confirms `coverage_warnings::check` is the only entry point, called from `typechecker.rs:5058`).

## Wins

- 0 heap allocations per warning emit (down from 2).
- Vec entry size shrinks from `(String, String)` (~48 bytes) to `(&'a str, &'static str)` (~32 bytes).
- Same diagnostic text and warning order.

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml --lib coverage_warnings` (7/7 green)

Closes #2348

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -463,6 +463,10 @@
     "resilient/src/audit_log_required.rs": {
       "branch": "res-2342-audit-log-short-circuit",
       "claimed_at": "2026-05-20T13:29:31Z"
+    },
+    "resilient/src/coverage_warnings.rs": {
+      "branch": "res-2350-coverage-warnings-borrow",
+      "claimed_at": "2026-05-20T14:19:39Z"
     }
   }
 }
--- a/resilient/src/coverage_warnings.rs
+++ b/resilient/src/coverage_warnings.rs
@@ -22,13 +22,20 @@
 
 use crate::Node;
 
+/// RES-2350: borrow `function` from the AST (`&'a str` of
+/// `Node::Function::name`) and use `&'static str` for `message` since
+/// every diagnostic in this module is a string literal. The previous
+/// `(String, String)` shape paid two heap allocations per warning —
+/// one to clone the fn name out of the AST, one to copy a static
+/// message literal into an owned `String` — for data that never
+/// escapes `check`.
 #[derive(Debug, Clone)]
-pub struct CoverageWarning {
-    pub function: String,
-    pub message: String,
+pub struct CoverageWarning<'a> {
+    pub function: &'a str,
+    pub message: &'static str,
 }
 
-pub fn analyze(program: &Node) -> Vec<CoverageWarning> {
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2351, opened 2026-05-20
RES-2352: modules::check — drop dead seen_order, short-circuit identifier walk

### LINKED ISSUE #2350: RES-2352: modules::check — drop dead seen_order Vec + short-circuit identifier walk
## Problem

Two small inefficiencies in `modules::check`:

1. **Dead `seen_order: Vec<&str>`** — built up via `push` on every module declaration but never read. Pure overhead (Vec allocation + per-module push).

2. **`visit` walks the full AST even after the first unresolved `mod::item` reference**. The closure short-circuits via `if unresolved.is_some() { return; }`, but `visit` still recurses through every remaining node — paying the recursive dispatch in `walk_children` for each one. RES-1238 added `any_node` for exactly this case.

\`\`\`rust
let mut unresolved: Option<String> = None;
crate::uniqueness_walk::visit(program, &mut |n| {
    if unresolved.is_some() { return; }  // ← still recurses, just a no-op closure
    if let Node::Identifier { name, .. } = n
        && let Some((mod_name, item_name)) = name.split_once("::")
        && let Some(exports) = module_items.get(mod_name)
        && !exports.contains(item_name)
    {
        unresolved = Some(format!("..."));
    }
});
\`\`\`

## Solution

- Delete `seen_order` and the per-module `push` (no callers).
- Replace `visit` with `any_node`, which propagates `true` upward as soon as `pred` matches, terminating the walk. `format!` produces the owned error message inside the closure, so no lifetime issues:

\`\`\`rust
let mut unresolved: Option<String> = None;
crate::uniqueness_walk::any_node(program, |n| {
    if let Node::Identifier { name, .. } = n
        && let Some((mod_name, item_name)) = name.split_once("::")
        && let Some(exports) = module_items.get(mod_name)
        && !exports.contains(item_name)
    {
        unresolved = Some(format!("{source_path}: error: module `{mod_name}` does not export `{item_name}`"));
        true
    } else {
        false
    }
});
if let Some(e) = unresolved { return Err(e); }
\`\`\`

## Wins

- Drops `seen_order` Vec allocation + per-module push.
- Short-circuits the second-pass identifier walk on the first unresolved reference instead of walking the entire program.
- Same diagnostic output (same error message, same first-match semantics — `visit` is pre-order, `any_node` is also pre-order).

## Acceptance criteria

- [ ] All 7 in-file tests pass (incl. `check_errors_on_unresolved_mod_item`, `check_ok_when_mod_item_exists`)

## Impact

`modules::check` runs when `markers.has_module_decl` is true (RES-1612 gate). On programs with many module declarations and many identifiers, the early-termination on first error is the standard pattern (RES-1238 / RES-2340 / RES-2342).

---

EricSpencer00:
## Summary

- Delete dead `seen_order: Vec<&str>` (built but never read).
- Replace `visit` with `any_node` in the second-pass identifier walk so it terminates on the first unresolved `mod::item` reference.

## Why

The previous `visit`-based shape had a `unresolved.is_some()` early-exit at the top of the closure, but `visit` still recursed through every remaining node — the closure became a no-op but `walk_children` dispatch fired for every sibling/aunt/descendant. `any_node` (RES-1238) propagates `true` upward and actually short-circuits the tree walk.

Same first-match semantics (both `visit` and `any_node` use pre-order traversal), same error message, fewer node visits on programs with at least one unresolved reference.

The `seen_order` Vec was orphaned by an earlier refactor — `module_items: HashMap<&str, HashSet<&str>>` already tracks declared modules and is the only consumer of that information.

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml --lib modules` (11/11 green incl. duplicate-module + unresolved-item tests)

Same pattern as RES-2340 (ghost_types), RES-2342 (audit_log_required).

Closes #2350

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -463,6 +463,10 @@
     "resilient/src/audit_log_required.rs": {
       "branch": "res-2342-audit-log-short-circuit",
       "claimed_at": "2026-05-20T13:29:31Z"
+    },
+    "resilient/src/modules.rs": {
+      "branch": "res-2352-modules-check-short-circuit",
+      "claimed_at": "2026-05-20T14:26:16Z"
     }
   }
 }
--- a/resilient/src/modules.rs
+++ b/resilient/src/modules.rs
@@ -88,7 +88,6 @@ pub(crate) fn check(program: &Node, source_path: &str) -> Result<(), String> {
     // First pass: collect module name → exported item names.
     let mut module_items: std::collections::HashMap<&str, std::collections::HashSet<&str>> =
         std::collections::HashMap::new();
-    let mut seen_order: Vec<&str> = Vec::new();
 
     for s in stmts {
         if let Node::ModuleDecl { name, body, .. } = &s.node {
@@ -100,7 +99,6 @@ pub(crate) fn check(program: &Node, source_path: &str) -> Result<(), String> {
                     source_path, n
                 ));
             }
-            seen_order.push(n);
             let mut exports: std::collections::HashSet<&str> = std::collections::HashSet::new();
             for item in body {
                 match item {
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2355, opened 2026-05-20
RES-2356: dce — replace O(n²) old-pc lookup with parallel new_to_old vec

### LINKED ISSUE #2354: RES-2356: dce — replace O(n²) old-pc lookup with parallel new_to_old vec
## Problem

Two relink passes in `dce.rs` walk `new_code` and, for each surviving jump op, do `(0..n).find(|&p| old_to_new[p] == new_pc)` to find the originating old PC. That's O(n) per jump, O(j × n) overall — quadratic in chunk size for chunks with many jumps.

\`\`\`rust
// remove_unreachable (line 116):
for (new_pc, op) in new_code.iter_mut().enumerate() {
    if !is_jump_op(*op) { continue; }
    let Some(old_pc) = (0..n).find(|&p| old_to_new[p] == new_pc) else { continue };
    ...
}

// fold_constant_branches (line 300):
for (new_pc, op) in new_code.iter_mut().enumerate() {
    if !is_jump_op(*op) { continue; }
    if replacement_jumps.iter().any(|(rp, _)| *rp == new_pc) { continue; }
    let Some(old_pc) = (0..chunk.code.len()).find(|&p| old_to_new[p] == new_pc) else { continue };
    ...
}
\`\`\`

The information is already known during the compaction loop — when we push to `new_code`, we already know the original `pc`. Tracking it in a parallel `new_to_old: Vec<usize>` makes the subsequent relink lookup O(1) instead of O(n).

## Solution

Build `new_to_old: Vec<usize>` alongside `new_code` during compaction. Then the relink pass just indexes:

\`\`\`rust
let old_pc = new_to_old[new_pc];
\`\`\`

Both `remove_unreachable` and `fold_constant_branches` use the same shape — both get the same treatment.

## Wins

- Drops the linear scan per surviving jump in both passes.
- Removes one O(j × n) term per DCE invocation. DCE runs once per compiled function (`compiler.rs:248` for fns + `compiler.rs:309` for main).
- No behavior change — `new_to_old[new_pc]` is exactly what the `find` returned before, just precomputed.

## Acceptance criteria

- [ ] All existing DCE tests pass (`ops_after_return_are_removed`, conditional-fold tests, etc.)

## Impact

DCE is called per-function during compilation. For programs with many functions and many jumps per function (typical for control-heavy embedded code), this O(n²) → O(n) change directly cuts compile time.

---

EricSpencer00:
## Summary

- Replace `(0..n).find(|&p| old_to_new[p] == new_pc)` linear scans in both `remove_unreachable` and `fold_constant_branches` with a precomputed `new_to_old: Vec<usize>`.
- O(j × n) → O(j + n) per DCE invocation (j = jumps, n = chunk size).

## Why

The relink pass walks `new_code` and, for each surviving jump op, needs to know the original PC. The previous shape recomputed that via a linear scan of `old_to_new`. But the information is already available during the compaction loop — when we push to `new_code`, we know the originating old `pc`.

Tracking a parallel `new_to_old: Vec<usize>` makes the lookup an array index.

## Wins

- Removes the quadratic relink term from both DCE rules.
- DCE runs per compiled function (`compiler.rs:248`) plus once for `main` (`compiler.rs:309`). On programs with many functions / many jumps per function, this dominates relink cost.
- No behavior change — `new_to_old[new_pc]` returns exactly what the `find` returned before.

## Mechanism

`remove_unreachable`:
- Push `pc` to `new_to_old` alongside `new_code.push(op)` inside the reachability loop.
- Relink: `let old_pc = new_to_old[new_pc];` replaces the `find`.

`fold_constant_branches`:
- Push `i` (the originating old PC) to `new_to_old` on both the verbatim-copy and `ReplaceWithJump` branches. The `RemoveBoth` branch pushes nothing (no new_code entry).
- For `ReplaceWithJump` entries, the entry maps back to the Const's old_pc — but the relink pass skips these entries via the existing `replacement_jumps.iter().any(...)` filter and re-targets them through `replacement_jumps` separately, so the back-pointer is unused for those slots. Kept consistent so the parallel-vec invariant `new_to_old.len() == new_code.len()` holds.

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml --lib dce` (12/12 green — reachability + constant-branch fold + line-info sync tests)

Closes #2354

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -463,6 +463,10 @@
     "resilient/src/audit_log_required.rs": {
       "branch": "res-2342-audit-log-short-circuit",
       "claimed_at": "2026-05-20T13:29:31Z"
+    },
+    "resilient/src/dce.rs": {
+      "branch": "res-2356-dce-relink-on-square",
+      "claimed_at": "2026-05-20T14:35:57Z"
     }
   }
 }
--- a/resilient/src/dce.rs
+++ b/resilient/src/dce.rs
@@ -88,6 +88,11 @@ fn remove_unreachable(chunk: &mut Chunk) {
     let mut old_to_new = vec![usize::MAX; n + 1];
     let mut new_code: Vec<Op> = Vec::with_capacity(n);
     let mut new_line_info: Vec<u32> = Vec::with_capacity(n);
+    // RES-2356: parallel `new_to_old[new_pc] = old_pc` lookup avoids
+    // the O(n) `(0..n).find(...)` per surviving jump in the relink
+    // pass. Building it during compaction is free; the lookup is
+    // then O(1).
+    let mut new_to_old: Vec<usize> = Vec::with_capacity(n);
 
     // Capture jump targets from original offsets BEFORE we mutate.
     let orig_targets: Vec<Option<usize>> = chunk
@@ -100,6 +105,7 @@ fn remove_unreachable(chunk: &mut Chunk) {
     for (pc, &op) in chunk.code.iter().enumerate() {
         if reachable[pc] {
             old_to_new[pc] = new_code.len();
+            new_to_old.push(pc);
             new_code.push(op);
             new_line_info.push(chunk.line_info[pc]);
         }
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2361, opened 2026-05-20
RES-2362: monotonic_field — tighten marker gate, drop redundant any_node pre-scan

### LINKED ISSUE #2360: RES-2362: monotonic_field — tighten marker gate to field assignments, drop redundant any_node pre-scan
## Problem

\`monotonic_field::check\` is gated in the typechecker by:

\`\`\`rust
if markers.any_fn_name_with_prefix(&[\"last_\", \"latest_\", \"max_\", \"monotonic_\"]) {
    crate::monotonic_field::check(program, source_path)?;
}
\`\`\`

But the pass itself only fires on \`Node::FieldAssignment\` nodes whose field name has a monotonic prefix/suffix. A monotonic-named *function* (e.g., \`fn last_seen_time(...)\`) triggers the call but produces no work — and inside, the pass already pre-scans via \`any_node\` to verify a monotonic FieldAssignment exists:

\`\`\`rust
let has_monotonic_assign = any_node(program, |n| match n {
    Node::FieldAssignment { field, .. } => is_monotonic_field(field),
    _ => false,
});
if !has_monotonic_assign { return Ok(()); }
\`\`\`

That \`any_node\` walk is redundant for programs that DO have such assignments (we proceed anyway) and is wasted work for programs that have a monotonic-named fn but no monotonic FieldAssignment (the common case for the false-positive trigger).

The codebase already exposes \`Markers::any_field_assigned_with_prefix_or_suffix(prefixes, suffixes)\` (pass_gate.rs:391) — exactly the tight gate this pass needs.

## Solution

Two coupled changes:

1. **typechecker.rs**: replace the loose \`any_fn_name_with_prefix\` gate with the precise \`any_field_assigned_with_prefix_or_suffix\` gate.

2. **monotonic_field.rs**: drop the now-redundant \`any_node\` pre-scan inside \`check\` — the typechecker marker guarantees the same condition.

\`\`\`rust
// typechecker.rs
if markers.any_field_assigned_with_prefix_or_suffix(
    &[\"last_\", \"latest_\", \"max_\", \"monotonic_\"],
    &[\"_seq\", \"_clock\", \"_epoch\", \"_tick\"],
) {
    crate::monotonic_field::check(program, source_path)?;
}
\`\`\`

\`\`\`rust
// monotonic_field.rs check():
// RES-2362: gated upstream by
// markers.any_field_assigned_with_prefix_or_suffix on MONO_PREFIXES /
// MONO_SUFFIXES — the previous internal any_node pre-scan was
// redundant.
let Node::Program(stmts) = program else { return Ok(()); };
for stmt in stmts { visit(&stmt.node, &mut |n| { ... }); }
\`\`\`

## Wins

- **Tighter gate**: programs with a \`last_*\` / \`max_*\` fn but no monotonic field assignment now skip the pass entirely (currently they enter \`check\`, run \`any_node\`, then bail).
- **Drops a redundant whole-AST walk** when the pass DOES need to run. The shared \`Markers::scan\` already populates \`field_names_assigned\`, so the new gate is O(1) lookup over that set.
- Matches the pattern used by \`audit_log_required\` (RES-2342), which uses the same gate.

## Acceptance criteria

- [ ] \`cargo test --manifest-path resilient/Cargo.toml\` green
- [ ] \`cargo clippy --all-targets -- -D warnings\` clean
- [ ] In-file tests (\`empty_program_returns_ok\`, \`program_without_trigger_returns_ok\`) pass
- [ ] No behavior change for programs that DO have monotonic FieldAssignments (existing diagnostics still fire)

## Impact

Removes one whole-AST walk per \`cargo build\` for the false-positive case (monotonic-named fn but no monotonic field assignment — a common shape since monotonic naming bleeds into helper functions). Removes the in-pass \`any_node\` walk for the true-positive case (the marker is already O(1)).

Same pattern as audit_log_required → markers gate elimination of internal pre-scans.

---

EricSpencer00:
## Summary

- Switch typechecker EXTENSION_PASSES gate from `any_fn_name_with_prefix` to `any_field_assigned_with_prefix_or_suffix` (the precise condition the pass actually depends on).
- Drop the now-redundant `any_node` pre-scan inside `monotonic_field::check`.

## Why

The pass only ever produces work for `Node::FieldAssignment` whose field name has a monotonic prefix/suffix. The previous gate fired on monotonic-named *functions*, which is a loose superset — programs with a helper like `fn last_seen_time(...)` entered the pass only to bail via an internal `any_node` walk.

`Markers::scan` already populates `field_names_assigned` once per compile, and `pass_gate.rs::any_field_assigned_with_prefix_or_suffix` (used by `audit_log_required`, RES-2342) does exactly the right check in O(set-size). The internal AST walk is pure redundancy after the marker gate tightens.

## Mechanism

\`\`\`rust
// typechecker.rs
if markers.any_field_assigned_with_prefix_or_suffix(
    &[\"last_\", \"latest_\", \"max_\", \"monotonic_\"],
    &[\"_seq\", \"_clock\", \"_epoch\", \"_tick\"],
) {
    crate::monotonic_field::check(program, source_path)?;
}
\`\`\`

`monotonic_field::check` no longer needs the `any_node` pre-scan (and drops the import).

## Wins

- Programs with monotonic-named helpers but no monotonic field assignment now skip the pass entirely.
- Programs that DO need the pass save one whole-AST walk; the new gate is O(`field_names_assigned.len()`) over a set the typechecker already builds.
- Same diagnostic output, same in-pass loop.

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml --lib monotonic_field` (2/2 green)

Same pattern as RES-2342 (`audit_log_required` gate).

Closes #2360

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -56,10 +56,6 @@
       "branch": "res-1256-secret-erasure-fast-reject",
       "claimed_at": "2026-05-12T03:39:06Z"
     },
-    "resilient/src/monotonic_field.rs": {
-      "branch": "res-1262-monotonic-fast-reject",
-      "claimed_at": "2026-05-12T03:53:13Z"
-    },
     "resilient/src/unsafe_check.rs": {
       "branch": "res-1281-unsafe-check-fast-reject",
       "claimed_at": "2026-05-12T04:21:23Z"
@@ -215,10 +211,6 @@
     "resilient/src/monomorph.rs": {
       "branch": "res-1924-monomorph-presize",
       "claimed_at": "2026-05-19T17:31:00Z"
-    },
-    "resilient/src/typechecker.rs": {
-      "branch": "res-1766-typechecker-hot-presize",
-      "claimed_at": "2026-05-13T11:55:37Z"
     },
     "resilient/src/supervisor.rs": {
       "branch": "res-1770-verifier-presize",
@@ -467,6 +459,14 @@
     "resilient/src/epoch_ordering.rs": {
       "branch": "res-2346-epoch-ordering-stream-prev",
       "claimed_at": "2026-05-20T13:58:05Z"
+    },
+    "resilient/src/monotonic_field.rs": {
+      "branch": "res-2362-monotonic-field-marker-gate",
+      "claimed_at": "2026-05-20T14:51:04Z"
+    },
+    "resilient/src/typechecker.rs": {
+      "branch": "res-2362-monotonic-field-marker-gate",
+      "claimed_at": "2026-05-20T14:51:04Z"
     }
   }
 }
--- a/resilient/src/monotonic_field.rs
+++ b/resilient/src/monotonic_field.rs
@@ -21,7 +21,7 @@
 )]
 
 use crate::Node;
-use crate::uniqueness_walk::{any_node, visit};
+use crate::uniqueness_walk::visit;
 
 const MONO_PREFIXES: &[&str] = &["last_", "latest_", "max_", "monotonic_"];
 const MONO_SUFFIXES: &[&str] = &["_seq", "_clock", "_epoch", "_tick"];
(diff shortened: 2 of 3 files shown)


EricSpencer00:

PR #2365, opened 2026-05-20
RES-2366: async_await + recursive_types — skip HashSet clone in is_X lookup

### LINKED ISSUE #2364: RES-2366: async_await + recursive_types — skip HashSet clone in is_async/is_recursive lookup
## Problem

\`async_await::is_async\` and \`recursive_types::is_recursive\` both have the same shape:

\`\`\`rust
pub fn is_async(name: &str) -> bool {
    ASYNC_FNS
        .read()
        .ok()
        .and_then(|g| g.clone())          // ← clones entire HashSet
        .map(|s| s.contains(name))
        .unwrap_or(false)
}
\`\`\`

\`g.clone()\` here is \`Option<HashSet<String>>::clone\` — it clones every entry in the set into a fresh \`HashSet\`, just to read it once. For a registry with N entries, every lookup pays N \`String::clone\` allocations.

The read guard is already held; \`HashSet::contains(&str)\` accepts a borrow via \`Borrow<str>\` on \`String\` keys. There's no need to materialize an owned copy.

## Solution

Hold the guard and borrow through the \`Option\`:

\`\`\`rust
pub fn is_async(name: &str) -> bool {
    ASYNC_FNS
        .read()
        .ok()
        .and_then(|g| g.as_ref().map(|s| s.contains(name)))
        .unwrap_or(false)
}
\`\`\`

\`g\` is the \`RwLockReadGuard\`; \`g.as_ref()\` auto-derefs through the guard to call \`Option::as_ref(&Option<HashSet<String>>)\` which gives \`Option<&HashSet<String>>\`. The bool result is extracted before \`g\` drops — no escaping borrow.

Same fix in:
- \`resilient/src/async_await.rs::is_async\`
- \`resilient/src/recursive_types.rs::is_recursive\`

\`iterator_protocol::is_iterator\` already uses this shape (\`g.contains(...)\` directly, since the value is a bare \`HashSet\` not \`Option<HashSet>\`).

## Wins

- Drops N \`String::clone\` per lookup (where N = registered async-/recursive-type count).
- Drops the per-call \`HashSet\` allocation.
- Same return value, same RwLock semantics.

## Acceptance criteria

- [ ] \`cargo test --manifest-path resilient/Cargo.toml\` green
- [ ] \`cargo clippy --all-targets -- -D warnings\` clean
- [ ] All in-file tests for both modules pass

## Impact

\`is_recursive\` and \`is_async\` are query APIs that the runtime/codegen layers consult for each struct type / fn name they inspect. The clone scales with how many \`#[recursive]\` / \`#[async]\` items are declared. On any project with multiple recursive types or async fns, this is hit per consultation.

Same pattern as the \`g.clone()\` → \`g.as_ref()\` lock-then-borrow refactor (RES-1544 / RES-1547 / RES-2006).

---

EricSpencer00:
## Summary

- Replace `.and_then(|g| g.clone()).map(|s| s.contains(name))` with `.and_then(|g| g.as_ref().map(|s| s.contains(name)))` in `async_await::is_async` and `recursive_types::is_recursive`.
- Drops the per-lookup `HashSet` clone (N `String` clones for N registered entries).

## Why

Both `is_X` functions held an `RwLock<Option<HashSet<String>>>`, took the read guard, then cloned the entire `Option<HashSet<String>>` just to call `.contains()`. The clone was pure overhead — `HashSet::contains` accepts `&str` via `Borrow<str>` on `String` keys, so the guard's borrow already suffices.

`iterator_protocol::is_iterator` already uses the borrow-through-guard shape (it holds a bare `HashSet` not wrapped in `Option`, but the pattern is identical).

## Mechanism

`g.as_ref()` auto-derefs through `RwLockReadGuard<'_, Option<HashSet<String>>>` to call `Option::as_ref(&Option<HashSet>)`, yielding `Option<&HashSet<String>>`. The bool result is extracted before the guard drops — no escaping borrow.

## Wins

- Drops N `String::clone` per `is_async` / `is_recursive` call.
- Drops the per-call `HashSet` allocation.
- Same return value, same lock semantics.

Same lock-then-borrow shape as RES-1544 / RES-1547 / RES-2006.

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml --lib 'async_await::'` (2/2 green)
- [x] `cargo test --manifest-path resilient/Cargo.toml --lib 'recursive_types::'` (5/5 green)

Closes #2364

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -467,6 +467,14 @@
     "resilient/src/epoch_ordering.rs": {
       "branch": "res-2346-epoch-ordering-stream-prev",
       "claimed_at": "2026-05-20T13:58:05Z"
+    },
+    "resilient/src/async_await.rs": {
+      "branch": "res-2366-is-x-rwlock-no-clone",
+      "claimed_at": "2026-05-20T15:16:39Z"
+    },
+    "resilient/src/recursive_types.rs": {
+      "branch": "res-2366-is-x-rwlock-no-clone",
+      "claimed_at": "2026-05-20T15:16:39Z"
     }
   }
 }
--- a/resilient/src/async_await.rs
+++ b/resilient/src/async_await.rs
@@ -36,11 +36,16 @@ pub fn install(set: HashSet<String>) {
 }
 
 pub fn is_async(name: &str) -> bool {
+    // RES-2366: hold the read guard and borrow through the
+    // `Option<HashSet<String>>` instead of cloning the whole set per
+    // lookup. `HashSet::contains` accepts a `&str` via `Borrow<str>`
+    // on `String` keys; the previous `g.clone()` materialized an
+    // owned copy of every entry for nothing. Same lock-then-borrow
+    // shape as RES-1544 / RES-1547 / RES-2006.
     ASYNC_FNS
         .read()
         .ok()
-        .and_then(|g| g.clone())
-        .map(|s| s.contains(name))
+        .and_then(|g| g.as_ref().map(|s| s.contains(name)))
         .unwrap_or(false)
 }
 
--- a/resilient/src/recursive_types.rs
+++ b/resilient/src/recursive_types.rs
@@ -30,11 +30,17 @@ pub fn install(set: HashSet<String>) {
 }
 
 pub fn is_recursive(type_name: &str) -> bool {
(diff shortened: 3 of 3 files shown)


EricSpencer00:

PR #2375, opened 2026-05-20
RES-2376: info_flow — filter inside walk_calls, drop non-secret callees

### LINKED ISSUE #2374: RES-2376: info_flow — filter inside walk_calls, drop intermediate non-secret callees
## Problem

`info_flow::check_program` collects every call-expression callee from a public fn's body into a `Vec<&str>`, then iterates the Vec and checks each against `secret_fns`:

\`\`\`rust
let mut leaks: Vec<&str> = Vec::new();
walk_calls(body, &mut leaks);  // pushes *every* callee name
for callee in leaks {
    if secret_fns.contains(callee) {
        errors.push(format!("info-flow: …"));
    }
}
\`\`\`

For each public fn we materialize a Vec of ALL its callees (including non-secret ones — typically the overwhelming majority) just to filter on the very next line. The intermediate storage is pure overhead.

(Note: an earlier draft of this fix was attempted as PR #2283 but didn't land — that branch is no longer in flight.)

## Solution

Move the filter inside the walker. `walk_calls` becomes a "find secret callees" walker that only pushes callees present in `secret_fns`. The caller's filter loop disappears:

\`\`\`rust
fn walk_calls<'a>(node: &'a Node, secret_fns: &HashSet<&str>, out: &mut Vec<&'a str>) {
    match node {
        Node::CallExpression { function, arguments, .. } => {
            if let Node::Identifier { name, .. } = function.as_ref()
                && secret_fns.contains(name.as_str())
            {
                out.push(name.as_str());
            }
            for a in arguments { walk_calls(a, secret_fns, out); }
        }
        // …
    }
}
\`\`\`

In `check_program`:

\`\`\`rust
let mut leaks: Vec<&str> = Vec::new();
walk_calls(body, &secret_fns, &mut leaks);
for callee in leaks {
    errors.push(format!("info-flow: `{name}` is `#[public]` but transitively calls `#[secret]` fn `{callee}`"));
}
\`\`\`

## Wins

- For typical public-fn bodies (many calls, few secrets), the Vec stops at the leak set — usually empty (no allocations) or small (one push per leak).
- The post-walk filter loop is eliminated.
- Same diagnostic output, same first-error semantics (`check` still returns `Err(errs[0])`).

## Acceptance criteria

- [ ] In-file tests pass (`secret_to_public_is_blocked`, `check_ok_without_attributes`, `check_program_no_attrs_returns_empty`)

## Impact

`info_flow::check` runs in the typechecker EXTENSION_PASSES block when at least one `#[public]`-tagged fn is declared. The wasted Vec storage scales with public-fn body size — the perf win is proportional to fn complexity.

---

EricSpencer00:
## Summary

- Move the `secret_fns.contains(callee)` filter from after the walk to inside `walk_calls`.
- The `leaks` Vec now stops at the actual leak set (usually empty), not every callee.

## Why

The previous shape pushed every callee identifier into `Vec<&str>`, then iterated and filtered against `secret_fns` on the very next line. For typical public-fn bodies (many calls, few secrets), the intermediate storage held data that was immediately discarded.

The walker already had access to the filter set conceptually — passing `&secret_fns` through is a one-parameter change that makes the walker emit only matching callees.

## Mechanism

\`\`\`rust
fn walk_calls<'a>(node: &'a Node, secret_fns: &HashSet<&str>, out: &mut Vec<&'a str>) {
    match node {
        Node::CallExpression { function, arguments, .. } => {
            if let Node::Identifier { name, .. } = function.as_ref()
                && secret_fns.contains(name.as_str())
            {
                out.push(name.as_str());
            }
            for a in arguments { walk_calls(a, secret_fns, out); }
        }
        // …
    }
}
\`\`\`

`check_program`'s emit loop drops the redundant filter — every entry in `leaks` is now known-secret.

## Wins

- For typical public-fn bodies, `leaks` stays empty → no Vec allocation.
- For bodies with N leaks (rare), `leaks` holds exactly N entries (was: every callee).
- Same diagnostic output, same first-error semantics.

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml --lib 'info_flow::'` (3/3 green)

(Note: an earlier draft was attempted as PR #2283 but the branch is no longer in flight.)

Closes #2374

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -471,6 +471,10 @@
     "resilient/src/peephole.rs": {
       "branch": "res-2368-peephole-relink-on-square",
       "claimed_at": "2026-05-20T15:20:44Z"
+    },
+    "resilient/src/info_flow.rs": {
+      "branch": "res-2376-info-flow-walk-filter",
+      "claimed_at": "2026-05-20T15:46:52Z"
     }
   }
 }
--- a/resilient/src/info_flow.rs
+++ b/resilient/src/info_flow.rs
@@ -67,16 +67,21 @@ pub fn check_program(program: &Node) -> Vec<String> {
     for s in stmts {
         if let Node::Function { name, body, .. } = &s.node {
             if public_fns.contains(name.as_str()) {
-                // walk body: any call to a secret-tagged fn is a leak.
+                // RES-2376: walk body collecting only secret callees.
+                // The previous shape pushed every callee identifier
+                // into `leaks: Vec<&str>` then filtered against
+                // `secret_fns` on the next line — for typical
+                // public-fn bodies (many calls, few secrets) the Vec
+                // held data that was immediately discarded. Filter
+                // inside the walk so `leaks` stops at the actual leak
+                // set (usually empty → no allocations).
                 let mut leaks: Vec<&str> = Vec::new();
-                walk_calls(body, &mut leaks);
+                walk_calls(body, &secret_fns, &mut leaks);
                 for callee in leaks {
-                    if secret_fns.contains(callee) {
-                        errors.push(format!(
-                            "info-flow: `{}` is `#[public]` but transitively calls `#[secret]` fn `{}`",
-                            name, callee
-                        ));
-                    }
+                    errors.push(format!(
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2391, opened 2026-05-20
RES-2392: dependent_arrays — drop dead DependentSpec::item_name field

### LINKED ISSUE #2390: RES-2392: dependent_arrays — drop dead DependentSpec::item_name field
## Problem

`DependentSpec` carries a `pub item_name: String` field set from the attribute item name. The two consumers in this module are:

1. `install()` at line 61: `g.insert(s.item_name.clone(), s)` — uses the field as HashMap key, then stores the spec (with the redundant name still inside).

2. `check()` at line 91: `specs.iter().find(|sp| sp.item_name == *name)` — linear-search by item_name.

Both can be expressed via `Vec<(String, DependentSpec)>` tuples without the field duplication. `DependentSpec` is `pub` but grep confirms zero external readers (only the typechecker's `dependent_arrays::check` is invoked from outside; `lookup`/`install` aren't used by non-test code).

## Solution

Drop `DependentSpec::item_name` and switch `collect()` to return `Vec<(String, DependentSpec)>`. Both consumers update:

- `install()`: `g.extend(specs)` — no per-item clone.
- `check()`'s line 91 find becomes `specs.iter().find(|(item, _)| item == name).map(|(_, sp)| sp)`.

## Wins

- Drops one `String::clone` per `#[dependent(...)]` at install.
- Drops 24 bytes of duplicated storage per registered spec.
- Aligns with the entire spec-registry family: wcet (RES-2190), prob (RES-2170), power (RES-2386), stack (RES-2388), phantom (RES-2390).

## Acceptance criteria

- [ ] All 5 in-file `dependent_arrays` tests pass (incl. `install_and_lookup_round_trip` which asserts `spec.item_name == "zip"` — needs test update)

---

EricSpencer00:
## Summary

- Drop `DependentSpec::item_name: String` and switch `collect()` to return `Vec<(String, DependentSpec)>`.
- Both consumers (`install()` extend, `check()` linear-find) update to operate on tuples.

## Why

`DependentSpec::item_name` duplicated the attribute key — used as the HashMap key in `install()` AND stored inside the value, and used as the linear-find key in `check()`. Same redundant-storage pattern as wcet (RES-2190), prob (RES-2170), power (RES-2386), stack (RES-2388), phantom (RES-2390).

## Wins

- Drops one `String::clone` per `#[dependent(...)]` at install.
- Drops 24 bytes of duplicated storage per registered spec.
- `DependentSpec` is `pub` but grep confirms zero external readers.

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml --lib 'dependent_arrays::'` (5/5 green)

In-file tests updated: `collects_length_var` now asserts `s[0].0 == "concat"` and `s[0].1.length_var == "M+N"`; `install_and_lookup_round_trip` drops the dead `spec.item_name` assertion.

Closes #2390

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,9 +1,5 @@
 {
   "claims": {
-    "resilient/src/distributed_invariants.rs": {
-      "branch": "res-1764-attr-collect-presize",
-      "claimed_at": "2026-05-13T11:45:18Z"
-    },
     "resilient/src/array_set_helpers.rs": {
       "branch": "res-2136-array-set-hashset-membership",
       "claimed_at": "2026-05-19T00:00:00Z"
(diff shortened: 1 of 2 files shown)


EricSpencer00:

PR #2393, opened 2026-05-20
RES-2394: mmio_regmap — drop duplicated MmioRegmap::struct_name field

### LINKED ISSUE #2392: RES-2394: mmio_regmap — drop duplicated MmioRegmap::struct_name field
## Problem

`MmioRegmap` carries `struct_name: String` set from the attribute item name. `install()` then uses it as the HashMap key:

\`\`\`rust
g.insert(m.struct_name.clone(), m);
\`\`\`

Every registered entry stores the struct name twice: once as the key, once as the value's field. The overlap check in `check()` reads `a.struct_name` / `b.struct_name` for the error message, but the strings already live in the (forthcoming) tuple position.

Same dead-field pattern as wcet (RES-2190), prob (RES-2170), power (RES-2386), stack (RES-2388), phantom (RES-2390), dependent (RES-2392).

## Solution

Drop `struct_name` from `MmioRegmap`, switch `collect()` to return `Vec<(String, MmioRegmap)>`, and have `install` use `g.extend(maps)`. The overlap check destructures the tuples for the error format.

`MmioRegmap` becomes `Copy` (two `u64`s). `lookup` returns `Option<MmioRegmap>` — the Copy makes `g.get(...).copied()` cheaper than the previous `.cloned()`.

`MmioRegmap` is `pub` but grep confirms zero external readers (only the typechecker's `mmio_regmap::check` is invoked from outside).

## Wins

- Drops one `String::clone` per `#[mmio(...)]` attribute at install.
- `MmioRegmap` shrinks from {String + 2×u64} to {2×u64} (24 bytes off the value heap).
- `lookup` uses `Copy` instead of String clone.

## Acceptance criteria

- [ ] In-file `mmio_regmap` tests pass

---

EricSpencer00:
## Summary

- Drop `MmioRegmap::struct_name: String` and switch `collect()` to return `Vec<(String, MmioRegmap)>`.
- `MmioRegmap` becomes `Copy` (two `u64`s); `install` uses `g.extend(maps)`; `lookup` switches to `.copied()`.

## Why

Same dead-field pattern as the wcet/prob/power/stack/phantom/dependent series. `install()` previously duplicated the name in both the HashMap key AND the value's field. The overlap-check error format reads the names from the tuple keys after the refactor — identical output.

`MmioRegmap` is `pub` but grep confirms zero external readers — only the typechecker's `mmio_regmap::check` is invoked from outside.

## Wins

- One `String::clone` dropped per `#[mmio(...)]` at install.
- `MmioRegmap` shrinks from {String + 2×u64} to {2×u64} — 24 bytes off the value heap.
- `lookup` returns a `Copy` value via `.copied()` instead of `.cloned()`.

## Test plan

- [x] `cargo test --manifest-path resilient/Cargo.toml --lib 'mmio_regmap::'` (2/2 green)

Closes #2392

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,9 +1,5 @@
 {
   "claims": {
-    "resilient/src/distributed_invariants.rs": {
-      "branch": "res-1764-attr-collect-presize",
-      "claimed_at": "2026-05-13T11:45:18Z"
-    },
     "resilient/src/array_set_helpers.rs": {
       "branch": "res-2136-array-set-hashset-membership",
       "claimed_at": "2026-05-19T00:00:00Z"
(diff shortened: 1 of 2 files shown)


EricSpencer00:

PR #2409, opened 2026-05-20
RES-2408: verifier_z3::extract_state_rhs — return borrowed &Node (drop clone)

### LINKED ISSUE #2408: RES-2408: verifier_z3::extract_state_rhs — return borrowed &Node instead of cloning
## Problem

`extract_state_rhs` (resilient/src/verifier_z3.rs:2712) returns `Result<Node, String>`. To produce that owned `Node`, the final arm of its match performs `Ok((**value).clone())` on the `Box<Node>` RHS of a `self.state = <int_expr>;` assignment.

Both callers in `check_pair_commute` (lines 2595 and 2603) immediately re-borrow the value via `&a_rhs` / `&b_rhs` when they pass it to `check_pair_commute_in(ctx, &a_rhs, &b_rhs, ...)` at line 2612. The owned `Node` is never consumed — it's used by reference only.

For an actor with N handlers, the commutativity verifier calls `check_pair_commute` over every unordered pair (N*(N-1)/2 calls), each paying two `Node::clone()` invocations on the handler body's RHS expression. The RHS can be a multi-level `InfixExpression` tree, so the cumulative AST-allocation cost in front of every commutativity check is real.

## Solution

Change the signature to `fn extract_state_rhs(body: &Node) -> Result<&Node, String>` and return `Ok(value.as_ref())` from the success arm. Update the two caller sites to drop the `&` they currently apply when forwarding into `check_pair_commute_in` (which already takes `&Node`).

The borrow chain works trivially — `body: &Node` outlives the returned `&Node`, and both callers use the result immediately within the same scope.

## Acceptance criteria

- [ ] `extract_state_rhs` returns `Result<&Node, String>` with no internal `clone`.
- [ ] Callers in `check_pair_commute` pass `a_rhs` / `b_rhs` (not `&a_rhs` / `&b_rhs`) to `check_pair_commute_in`.
- [ ] `cargo build` + `cargo test` (non-z3 build) green.
- [ ] CI `cargo test --features z3` covers the commutativity verifier paths.

## Impact

The Z3 commutativity verifier (`verify_commutative_handlers`) runs once per `actor` declaration that includes `commutative_with` annotations. Each pair check pays two of these clones; dropping both removes the AST-allocation overhead in the wrapper around the Z3 solver call. Same general pattern as RES-2406 (cluster_verifier) — lift clones that exist purely for ownership reasons when the value is only consumed by reference.

---

EricSpencer00:
## Summary

Closes #2408.

`extract_state_rhs` (resilient/src/verifier_z3.rs:2712) cloned the RHS expression of a `self.state = <int_expr>;` assignment via `Ok((**value).clone())` purely to satisfy a `Result<Node, String>` signature. Both callers in `check_pair_commute` (lines 2595, 2603) immediately re-borrowed the result via `&a_rhs` / `&b_rhs` when passing it into `check_pair_commute_in(ctx, &Node, &Node, ...)` at line 2612 — the owned `Node` was never consumed.

Changed the signature to `Result<&Node, String>`, returned `value.as_ref()`, and dropped the `&` on the forwarded arguments at the call sites.

## Test plan

- [x] `cargo check --features z3 --lib` green.
- [x] `cargo clippy --features z3 --lib -- -D warnings` clean.
- [x] `cargo build --manifest-path resilient/Cargo.toml` (non-z3) green.
- [ ] CI `cargo test --features z3` covers the commutativity verifier paths.

Pure refactor — no behaviour change. Each handler pair check inside the commutativity verifier now skips two deep AST clones; for an actor with N commutative-annotated handlers, that saves 2 clones × N*(N-1)/2 pair checks.

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -272,10 +272,6 @@
       "branch": "res-1901-hof-clone-elim",
       "claimed_at": "2026-05-19T05:02:34Z"
     },
-    "resilient/src/verifier_z3.rs": {
-      "branch": "res-1920-collect-cert-idents-single-walk",
-      "claimed_at": "2026-05-19T17:06:58Z"
-    },
     "resilient/src/compiler.rs": {
       "branch": "res-1922-compile-match-presize",
       "claimed_at": "2026-05-19T17:22:52Z"
@@ -463,6 +459,10 @@
     "resilient/src/phantom_types.rs": {
       "branch": "res-2390-phantom-types-drop-type-name",
       "claimed_at": "2026-05-20T16:19:52Z"
+    },
+    "resilient/src/verifier_z3.rs": {
+      "branch": "res-2408-extract-state-rhs-borrow",
+      "claimed_at": "2026-05-20T17:28:22Z"
     }
   }
 }
--- a/resilient/src/verifier_z3.rs
+++ b/resilient/src/verifier_z3.rs
@@ -2609,7 +2609,7 @@ fn check_pair_commute(a: &ActorHandler, b: &ActorHandler) -> CommutativityResult
         }
     };
 
-    Z3_CTX.with(|ctx| check_pair_commute_in(ctx, &a_rhs, &b_rhs, &a.name, &b.name))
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2411, opened 2026-05-20
RES-2410: region_inference::resolve — borrow input, skip per-link Region clones

### LINKED ISSUE #2410: RES-2410: region_inference::RegionTable::resolve — borrow input, skip per-link Region clones
## Problem

`RegionTable::resolve` (resilient/src/region_inference.rs:66) walks a union-find chain by repeatedly cloning each intermediate `Region`:

```rust
pub fn resolve(&self, mut r: Region) -> Region {
    loop {
        match &r {
            Region::Var(v) => match self.parent.get(&v.0) {
                Some(parent) => r = parent.clone(),  // ← clone per link
                None => return r,
            },
            Region::Named(_) => return r,
        }
    }
}
```

Because `Region::Named` carries a `String`, every intermediate clone allocates. A chain of length N pays N-1 `Region::Named` String allocations (or zero for `Region::Var` links) per resolve call.

The two `get_resolved` / `get_local_resolved` accessors (lines 161, 167) also `.clone()` the borrowed `Region` from the map just to satisfy the by-value signature:

```rust
self.entries.get(key).map(|r| self.table.resolve(r.clone()))
```

That clone is pure plumbing — `resolve` only ever reads its input.

## Solution

Change `resolve` to borrow its input:

```rust
pub fn resolve(&self, r: &Region) -> Region {
    let mut cur = r;
    loop {
        match cur {
            Region::Var(v) => match self.parent.get(&v.0) {
                Some(parent) => cur = parent,  // borrow, no clone
                None => return cur.clone(),
            },
            Region::Named(_) => return cur.clone(),
        }
    }
}
```

Then drop the `.clone()` on the two map accessor sites and pass `&a`, `&b` from `unify`. The function still returns owned `Region` (callers like `unify` need that), but the chain walk eliminates all intermediate allocations — only the final representative is cloned.

## Acceptance criteria

- [ ] `resolve` takes `&Region`, walks the chain by reference, and clones only the final representative.
- [ ] `get_resolved` and `get_local_resolved` drop their `r.clone()`.
- [ ] `unify` updated to pass `&a` and `&b` to `resolve`.
- [ ] Existing tests pass (chain-of-three, conflict, idempotent unify).

## Impact

Region inference runs during the typechecker pass on any program that uses `&'region` borrow types. Every `unify` call invokes `resolve` twice; every typechecker post-pass on a region-annotated parameter invokes `get_resolved`. For chains formed by transitive `unify(A, B); unify(B, C);` patterns, the per-link allocation savings compound. Same general pattern as RES-2406 / RES-2408: drop clones that exist purely to satisfy a by-value signature when the function reads its input by reference.

---

EricSpencer00:
## Summary

Closes #2410.

`RegionTable::resolve` (resilient/src/region_inference.rs:66) walked the union-find chain by reassigning `r = parent.clone()` at every hop. `Region::Named` carries a `String`, so each intermediate clone allocated; a chain of length N paid up to N-1 unnecessary `String` allocations.

Changed `resolve` to take `&Region`, walk by reference into `self.parent`, and clone only the final representative (which is what callers like `unify`'s by-value match arms still need). Dropped the `.clone()` plumbing on the two `get_resolved` / `get_local_resolved` map accessors and on `unify`'s two `self.resolve(...)` call sites.

## Test plan

- [x] `cargo build --manifest-path resilient/Cargo.toml` green.
- [x] All 15 `region_inference::tests` pass — chain-of-three, conflict detection, idempotent unify, infer_region_subst conflict, build_region_map collects.
- [x] Three test call sites updated to pass `&Region::Var(v)`.

External API unchanged: `get_resolved` / `get_local_resolved` still return `Option<Region>`; lib.rs callers pass `&ParamKey` exactly as before.

Pure refactor — no behaviour change. Same pattern as RES-2406 (cluster_verifier) and RES-2408 (verifier_z3::extract_state_rhs): drop clones that exist purely to satisfy a by-value signature when the function only reads its input.

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -156,10 +156,6 @@
       "branch": "res-2006-semver-baseline-borrow",
       "claimed_at": "2026-05-19T00:00:00Z"
     },
-    "resilient/src/region_inference.rs": {
-      "branch": "res-2146-region-inference-callee-info-borrow",
-      "claimed_at": "2026-05-19T00:00:00Z"
-    },
     "resilient/src/mutual_recursion_scc.rs": {
       "branch": "res-2086-mutual-recursion-scc-borrow",
       "claimed_at": "2026-05-19T00:00:00Z"
@@ -463,6 +459,10 @@
     "resilient/src/phantom_types.rs": {
       "branch": "res-2390-phantom-types-drop-type-name",
       "claimed_at": "2026-05-20T16:19:52Z"
+    },
+    "resilient/src/region_inference.rs": {
+      "branch": "res-2410-region-resolve-borrow",
+      "claimed_at": "2026-05-20T17:33:41Z"
     }
   }
 }
--- a/resilient/src/region_inference.rs
+++ b/resilient/src/region_inference.rs
@@ -63,14 +63,23 @@ impl RegionTable {
     ///
     /// Follows variable chains until a `Region::Named` or an unbound
     /// `Region::Var` is reached.
-    pub fn resolve(&self, mut r: Region) -> Region {
+    ///
+    /// RES-2410: input is borrowed and intermediate parent links are
+    /// followed by reference into `self.parent`. Only the final
+    /// representative is cloned (callers like `unify`'s match arms
+    /// need an owned `Region`). The previous shape paid one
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2415, opened 2026-05-20
RES-2414: linear — skip redundant bindings snap clones in If/Match arm walks

### LINKED ISSUE #2414: RES-2414: linear — skip redundant bindings snap clones in If/Match arm walks
## Problem

`linear::walk` (resilient/src/linear.rs) has two `bindings.clone()` patterns that pay for restores that are already covered by later moves:

### Non-z3 If arm (lines 286–298)

```rust
let snap = bindings.clone();
walk(consequence, bindings, fn_name, source_path)?;
*bindings = snap.clone();                  // ← restore via clone
if let Some(alt) = alternative {
    walk(alt, bindings, fn_name, source_path)?;
}
*bindings = snap;                           // ← final restore via move
```

When `alternative` is `None`, the alt-walk is skipped and the trailing `*bindings = snap;` overwrites whatever the `snap.clone()` produced — the intermediate clone is dead weight. Same exact shape as the RES-1386 fix in `verifier_loop_invariants::walk`.

### Match arm (lines 317–330)

```rust
let snap = bindings.clone();
for (_, guard, body) in arms {
    *bindings = snap.clone();              // ← always cloned, even on first iter
    if let Some(g) = guard { walk(g, ...); }
    walk(body, ...);
}
*bindings = snap;
```

On the **first** iteration, `bindings` already equals `snap` (the clone above just initialised snap from current bindings), so the `snap.clone()` is redundant — it would assign bindings the same value it already has. Subsequent iterations genuinely need the restore because the previous iter's walk mutated bindings.

## Solution

1. Reorder the non-z3 If arm so the snap restore happens **inside** the `if let Some(alt)` branch:
   ```rust
   let snap = bindings.clone();
   walk(consequence, bindings, fn_name, source_path)?;
   if let Some(alt) = alternative {
       *bindings = snap.clone();
       walk(alt, bindings, fn_name, source_path)?;
   }
   *bindings = snap;
   ```
2. Peel the first Match arm out of the loop, or gate the `*bindings = snap.clone()` with an `is_first` flag, so the first arm uses the already-equal `bindings` and only subsequent arms pay the restore clone.

## Acceptance criteria

- [ ] Non-z3 If-without-else paths skip the intermediate `snap.clone()`.
- [ ] First Match arm doesn't pay a per-iter `snap.clone()`.
- [ ] All `linear::tests` pass — consumes-twice rejection, double-consume in arm, use-after-consume.

## Impact

`linear::walk` runs as a typechecker pass on any program containing `#[linear]`-typed bindings. Every if-without-else in such a program previously paid one HashMap clone per call; every match paid `arms.len()` clones instead of `arms.len() - 1`. The bindings map carries `String` keys + `LinearBinding` values, so cloning isn't free. Same general pattern as RES-1386 (verifier_loop_invariants) — drop snapshot clones whose effect is always overwritten before being observed.

---

EricSpencer00:
## Summary

Closes #2414.

Two `bindings.clone()` calls in `linear::walk` paid for restores that were already covered by later moves.

**Non-z3 If arm (lines 286–298):** when `alternative` is `None`, the trailing `*bindings = snap;` already overwrote whatever `*bindings = snap.clone();` had produced. Gated the intermediate clone on `alternative.is_some()` — same shape as RES-1386's fix in `verifier_loop_invariants::walk`.

**Match arm (lines 317–330):** on the first arm iteration, `bindings` already equals `snap` (just cloned from current bindings), so the leading `*bindings = snap.clone();` was assigning bindings a fresh copy of a value it already held. Peeled the first iter via an `is_first` gate so only iters 2..N pay the restore clone — savings of `arms.len() - 1` instead of the previous `arms.len()`.

## Test plan

- [x] `cargo build` green.
- [x] All 52 linear-related tests pass: consumes-twice rejection, double-consume in arm, deep linear recursion runtime-error path, linear_value_used_once_is_accepted, etc.

`linear::walk` runs as a typechecker pass on any program with `#[linear]`-typed bindings. The bindings HashMap carries `String` keys and `LinearBinding` values — neither side is free to clone.

Pure refactor — no behaviour change. Same pattern as RES-1386: drop snapshot clones whose effect is always overwritten before being observed.

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -463,6 +463,10 @@
     "resilient/src/row_polymorphism.rs": {
       "branch": "res-2398-row-polymorphism-drop-fn-name",
       "claimed_at": "2026-05-20T16:41:54Z"
+    },
+    "resilient/src/linear.rs": {
+      "branch": "res-2414-linear-snap-clones",
+      "claimed_at": "2026-05-20T17:44:54Z"
     }
   }
 }
--- a/resilient/src/linear.rs
+++ b/resilient/src/linear.rs
@@ -288,10 +288,18 @@ fn walk(
                 // Without Z3, use the conservative snapshot/restore
                 // approach: each branch starts fresh, consumption
                 // in one branch doesn't affect the merge.
+                //
+                // RES-2414: gate the inter-branch `snap.clone()` on
+                // there actually being an alternative branch. When
+                // `alternative` is `None`, the trailing
+                // `*bindings = snap;` already overwrites whatever
+                // bindings ended up as after `consequence`, so the
+                // intermediate clone was pure waste. Same shape as
+                // RES-1386's fix in `verifier_loop_invariants`.
                 let snap = bindings.clone();
                 walk(consequence, bindings, fn_name, source_path)?;
-                *bindings = snap.clone();
                 if let Some(alt) = alternative {
+                    *bindings = snap.clone();
                     walk(alt, bindings, fn_name, source_path)?;
                 }
                 *bindings = snap;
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2431, opened 2026-05-20
RES-2430: lock_ordering — nested pair_sites map drops per-iter (b,a) tuple clone

### LINKED ISSUE #2430: RES-2430: lock_ordering — nested pair_sites map drops per-iter (b,a) tuple clone
## Problem

`lock_ordering::check` (resilient/src/lock_ordering.rs:36) stores co-acquisition pair-sites in a flat `HashMap<(String, String), Vec<String>>` keyed on `(prior_lock, current_lock)`. The inversion-detection loop then reverses the key and probes for the inverted pair:

```rust
let mut pair_sites: HashMap<(String, String), Vec<String>> = HashMap::new();
...
for ((a, b), fns_ab) in &pair_sites {
    let key_ba = (b.clone(), a.clone());          // ← 2 String clones per iter
    if let Some(fns_ba) = pair_sites.get(&key_ba) {
        ...
    }
}
```

Two String clones per pair_sites entry, even though `pair_sites` already owns both strings. The comment at lines 50-56 claims `(&str, &str)` would work via `Borrow` on tuple-of-borrows — that's incorrect; stdlib has no `Borrow<(&str, &str)>` impl for `(String, String)`. So the flat-key shape genuinely forces the owned-tuple lookup.

## Solution

Re-shape `pair_sites` as nested `HashMap<String, HashMap<String, Vec<String>>>` (lock-`prior` → set of `current` → function-name list). The inversion check becomes:

```rust
for (a, by_b) in &pair_sites {
    for (b, fns_ab) in by_b {
        if let Some(fns_ba) = pair_sites.get(b).and_then(|m| m.get(a)) {
            ...
        }
    }
}
```

Both lookups use the existing `String: Borrow<str>` impl — no clone, no tuple construction. Same general principle as RES-2008 / RES-2010 / RES-2012 / RES-2014 / RES-2184 / RES-2194 (nested-map shape replacing flat tuple keys) and RES-2418 (lint allow comments line-keyed map).

`collect_pairs` updated to populate the nested shape:

```rust
pairs.entry(prior.clone()).or_default()
    .entry(lk.clone()).or_default()
    .push(fn_name.to_string());
```

Same total clone count as the flat shape (two `String::clone` per pair-insertion via the two `entry()` calls), so the build cost is unchanged. The savings are exclusively in the inversion-check loop.

## Acceptance criteria

- [ ] `pair_sites: HashMap<String, HashMap<String, Vec<String>>>`.
- [ ] Inversion check uses chained `.get(...)` lookups — no `key_ba.clone()` plumbing.
- [ ] Stale "tuple-of-borrows works via Borrow" comment removed.
- [ ] Existing tests pass (empty program + no-trigger + integration via the broader suite).

## Impact

The pass runs whenever a program contains `lock`/`acquire`/`mutex_lock` calls or `lock_*` / `unlock_*` prefixed calls. For each unique `(a, b)` pair in `pair_sites`, the previous shape allocated 2 transient `String`s for the reversed-key lookup. For K unique pairs, that's 2K avoidable allocations per pass. Nested-map shape also matches the existing pattern used elsewhere in the codebase (RES-2008/2010/2012/2014).

---

EricSpencer00:
## Summary

Closes #2430.

`lock_ordering::check` stored co-acquisition pairs in a flat `HashMap<(String, String), Vec<String>>`. The inversion-detection loop allocated two transient `String`s per entry just to construct the reversed key for lookup — stdlib has no `Borrow<(&str, &str)>` impl for `(String, String)`, so the flat-key shape genuinely required the owned-tuple probe.

Re-shaped `pair_sites` as `HashMap<String, HashMap<String, Vec<String>>>` (prior → current → fn_names). The inversion loop chains `.get(b).and_then(|m| m.get(a))` — both lookups use the existing `String: Borrow<str>` impl, zero allocations.

`collect_pairs` builds the nested shape via chained `entry().or_default()` calls — same two clones per insertion as the flat-key version, so the build cost is unchanged. Savings are exclusively in the inversion-check: K unique pairs save 2K transient allocations.

## Test plan

- [x] `cargo build` green.
- [x] Both `lock_ordering::tests` pass.

Same nested-map shape pattern as RES-2008 / RES-2010 / RES-2012 / RES-2014 / RES-2184 / RES-2194 / RES-2418.

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -112,10 +112,6 @@
       "branch": "res-1538-termination-fn-info-borrow",
       "claimed_at": "2026-05-12T15:15:14Z"
     },
-    "resilient/src/lock_ordering.rs": {
-      "branch": "res-1752-lock-ordering-presize",
-      "claimed_at": "2026-05-13T11:18:18Z"
-    },
     "resilient/src/power_contracts.rs": {
       "branch": "res-2018-attr-move-item",
       "claimed_at": "2026-05-19T00:00:00Z"
@@ -212,10 +208,6 @@
       "branch": "res-2014-assoc-const-nested-map",
       "claimed_at": "2026-05-19T00:00:00Z"
     },
-    "resilient/src/rate_limit_static.rs": {
-      "branch": "res-1962-reentrancy-rate-limit-presize",
-      "claimed_at": "2026-05-19T18:59:39Z"
-    },
     "resilient/src/sum_types.rs": {
       "branch": "res-1788-rate-limit-presize",
       "claimed_at": "2026-05-13T12:35:40Z"
@@ -303,10 +295,6 @@
     "resilient/src/assume_axioms.rs": {
       "branch": "res-1958-assume-axioms-presize",
       "claimed_at": "2026-05-19T18:52:28Z"
-    },
-    "resilient/src/reentrancy_guard.rs": {
-      "branch": "res-1962-reentrancy-rate-limit-presize",
-      "claimed_at": "2026-05-19T18:59:39Z"
     },
     "resilient/src/array_binary_search.rs": {
       "branch": "res-2034-binary-search-no-intermediate-vec",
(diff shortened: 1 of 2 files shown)


EricSpencer00:

PR #2481, opened 2026-05-25
RES-2498: remove unsound Neg;Neg peephole fold

### LINKED ISSUE #2480: RES-2498: peephole Neg;Neg fold is unsound under Saturate/Trap overflow modes
## Problem

Peephole Rule 17 (`Neg; Neg → identity`) drops both negation operations as a no-op, claiming `--x == x` for all integers. This is only true in **Wrap** overflow mode.

In **Saturate** mode:
- `x = i64::MIN`
- First `Neg`: `saturating_neg(i64::MIN)` = `i64::MAX`
- Second `Neg`: `saturating_neg(i64::MAX)` = `-i64::MAX` = `i64::MIN + 1`
- Expected: `i64::MIN + 1`
- After peephole: `i64::MIN` ← **wrong by 1**

In **Trap** mode:
- `x = i64::MIN`
- First `Neg`: overflow error (i64::MIN has no positive counterpart)
- After peephole: no error, `y = i64::MIN` ← **silent miscompilation**

## Solution

The fix is straightforward: `rule_double_neg` must be conditioned on the overflow mode being `Wrap`. Pass the `OverflowMode` into the peephole optimizer and guard the rule.

Alternatively, the simplest and safest fix is to **remove Rule 17 entirely**. The optimization saves one instruction in a rarely-occurring pattern (`--x`); the correctness risk is high. The value is negligible compared to the semantic danger.

## Acceptance criteria
- [ ] `Neg; Neg` is NOT folded when overflow mode is `Saturate` or `Trap`
- [ ] Test: `--i64::MIN` in Trap mode produces an overflow error
- [ ] Test: `--i64::MIN` in Saturate mode produces `i64::MIN + 1`, not `i64::MIN`

## Impact
Tier 1 compiler correctness — miscompilation that silently produces wrong results or suppresses overflow errors, depending on overflow mode.

---

EricSpencer00:
## Summary
- Removes peephole Rule 17 (`Neg; Neg → identity`) which was unsound under Saturate and Trap overflow modes
- In Saturate mode: `--i64::MIN` = `i64::MIN + 1`, not `i64::MIN` — the fold produced a wrong result
- In Trap mode: first `Neg` on `i64::MIN` should error, but the fold silently suppressed the overflow

## Why removal over conditioning
The optimization saves one instruction in a rarely-occurring pattern (`--x`). Threading overflow mode through the peephole API adds complexity for negligible benefit. Removing the rule is simpler and eliminates the risk entirely.

## Test changes
- Removed 5 tests for the old `rule_double_neg` function (function no longer exists)
- Added `rule17_removed_neg_neg_not_folded` which verifies `Neg; Neg` survives the optimizer unchanged

Closes #2480

---

### DIFF
--- a/resilient/src/peephole.rs
+++ b/resilient/src/peephole.rs
@@ -28,7 +28,7 @@
 //! 14. `Const(k==0); Sub`                   → drop both (x - 0 == x, identity)
 //! 15. `Const(k==1); Div`                   → drop both (x / 1 == x, identity)
 //! 16. `Not; Not`                            → drop both (!!b == b, idempotent)
-//! 17. `Neg; Neg`                            → drop both (--x == x, double int-neg identity)
+//! 17. (REMOVED — RES-2498: unsound under Saturate/Trap overflow modes)
 //! 18. `Not; JumpIfTrue(off)`               → `JumpIfFalse(off)` (mirror of Rule 4)
 //! 19. `LoadLocal(n); StoreLocal(n)`        → drop both (self-load-store is a no-op)
 //! 20. `Const(true); JumpIfTrue(off)`       → `Jump(off)` (branch always taken)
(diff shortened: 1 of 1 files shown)


EricSpencer00:

PR #2483, opened 2026-05-25
RES-2500: fix const_fold overflow soundness under Saturate/Trap modes

### LINKED ISSUE #2482: RES-2500: const_fold uses wrapping arithmetic, unsound under Saturate/Trap overflow
## Problem

The constant-fold optimizer (`const_fold.rs`) evaluates compile-time arithmetic using `wrapping_add`, `wrapping_sub`, `wrapping_mul`, and `wrapping_neg`. This produces results that silently differ from what the VM computes at runtime when the overflow mode is Saturate or Trap.

Example: `let x = 9223372036854775807 + 1;`
- Wrap mode: `i64::MIN` (correct match)
- Trap mode: runtime overflow error, but const_fold produces `i64::MIN`
- Saturate mode: `i64::MAX`, but const_fold produces `i64::MIN`

Similarly for `try_fold_unop`: `Const(i64::MIN); Neg` is folded to `Const(i64::MIN)` via `wrapping_neg`, but Trap mode should error and Saturate mode should produce `i64::MAX`.

Same class of bug as RES-2498 (peephole `Neg;Neg` fold).

## Solution

Guard arithmetic folds with overflow checks. Only fold when the result is the same under all overflow modes — i.e., when no overflow occurs. Use `checked_add/sub/mul/neg` and skip the fold when `None` (overflow detected).

This is safe and mode-independent: if an operation doesn't overflow, all three modes produce the same result. If it does overflow, the runtime should handle it per the configured mode.

## Acceptance criteria
- [ ] `Const(a); Const(b); Add` is NOT folded when `a.checked_add(b).is_none()`
- [ ] Same for Sub, Mul
- [ ] `Const(a); Neg` is NOT folded when `a.checked_neg().is_none()`
- [ ] Test: `Const(i64::MAX); Const(1); Add` is NOT folded
- [ ] Test: `Const(i64::MIN); Neg` is NOT folded
- [ ] Non-overflowing folds still fire (regression test)

## Impact
Tier 1 compiler correctness — silent miscompilation. The compiled program produces different results than a runtime-interpreted version when overflow occurs.

---

EricSpencer00:
## Summary
- Replaces `wrapping_add/sub/mul/neg` with `checked_add/sub/mul/neg` in the constant-fold optimizer
- Overflowing operations are now left unfolded, letting the VM handle them per the configured overflow mode
- Non-overflowing folds continue to fire — no regression in optimization coverage for normal code

## Why this matters
Same class of bug as RES-2498 (peephole `Neg;Neg` fold). The constant folder was silently miscompiling code when overflow occurs:
- `i64::MAX + 1` folded to `i64::MIN` (wrapping) instead of `i64::MAX` (saturate) or error (trap)
- `-(i64::MIN)` folded to `i64::MIN` (wrapping_neg) instead of `i64::MAX` (saturate) or error (trap)

For a safety-critical compiler, compile-time evaluation must agree with runtime semantics.

## Test changes
- `folds_int_add_with_overflow_wraps` → renamed to `does_not_fold_int_add_with_overflow`: previously verified wrapping result, now verifies the fold does NOT fire
- Added `does_not_fold_neg_on_int_min`, `does_not_fold_sub_with_overflow`, `does_not_fold_mul_with_overflow`

Closes #2482

---

### DIFF
--- a/resilient/src/const_fold.rs
+++ b/resilient/src/const_fold.rs
@@ -14,9 +14,9 @@
 //!
 //! | Pattern | Result |
 //! |---|---|
-//! | `Const(i); Const(j); Add` | `Const(i + j)` (wrapping) |
-//! | `Const(i); Const(j); Sub` | `Const(i - j)` (wrapping) |
-//! | `Const(i); Const(j); Mul` | `Const(i * j)` (wrapping) |
+//! | `Const(i); Const(j); Add` | `Const(i + j)` — skipped on overflow |
+//! | `Const(i); Const(j); Sub` | `Const(i - j)` — skipped on overflow |
+//! | `Const(i); Const(j); Mul` | `Const(i * j)` — skipped on overflow |
 //! | `Const(i); Const(j); Div` | `Const(i / j)` — skipped if `j == 0` |
 //! | `Const(i); Const(j); Mod` | `Const(i % j)` — skipped if `j == 0` |
 //! | `Const(i); Const(j); Eq`/`Neq`/`Lt`/`Le`/`Gt`/`Ge` | `Const(bool)` |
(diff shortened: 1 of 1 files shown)


EricSpencer00:

PR #2485, opened 2026-05-25
RES-2502: compile labeled break/continue to bytecode

### LINKED ISSUE #2484: RES-2502: compile labeled break/continue to bytecode
## Problem

The parser creates `Node::BreakLabel` and `Node::ContinueLabel` AST nodes, and the tree-walker interpreter handles them correctly via `Value::BreakLabel`/`Value::ContinueLabel` sentinel values. However, the bytecode compiler (`compiler.rs`) does **not** handle these nodes — they fall through to `Err(CompileError::Unsupported(...))`.

This means labeled break/continue (`break outer;`, `continue outer;`) works in the tree-walker but silently fails in the VM path, creating a semantic divergence between the two execution engines.

## Solution

1. Add a `label: Option<String>` field to `LoopState`
2. Replace the single `Option<&mut LoopState>` parameter with `&mut Vec<LoopState>` (a loop stack) so nested loops can all be accessed by label
3. Add `Node::BreakLabel` and `Node::ContinueLabel` dispatch in both `compile_stmt` and `compile_stmt_in_fn`
4. When entering a while/for-in loop, push a LoopState (with label if present) onto the stack; pop and patch after body compilation
5. `break label` searches the stack for the matching label and patches to that loop's exit
6. `continue label` searches the stack for the matching label and patches to that loop's continue target

## Acceptance criteria

- [ ] `break outer;` from a nested for-in/while loop compiles and produces correct results via the VM
- [ ] `continue outer;` from a nested loop compiles and produces correct results via the VM
- [ ] Unlabeled `break;` and `continue;` remain unchanged in behavior
- [ ] `break nonexistent_label;` produces a clean compile error
- [ ] All existing tests pass
- [ ] New VM-path tests cover labeled break/continue

## Impact

Closes a tier 1 semantic divergence between the tree-walker and VM execution paths. Labeled break/continue is the only parsed feature that works in the tree-walker but not the VM.

---

EricSpencer00:
## Summary

- Labeled `break`/`continue` (`break outer;`, `continue outer;`) was parsed and handled by the tree-walker but **not compiled to bytecode** — the compiler fell through to `Err(CompileError::Unsupported(...))`, creating a semantic divergence between the two execution engines
- Replaced the single `Option<&mut LoopState>` parameter threading with a `Vec<LoopState>` stack so labeled break/continue can target any enclosing loop by name
- Each while/for-in loop pushes a `LoopState` (with optional label from the AST node) onto the stack; the entry is popped and its break/continue patches are resolved after the body is compiled
- Unlabeled break/continue targets the last (innermost) stack entry; labeled variants search the stack in reverse for a matching label

## Test plan

- [x] 7 new VM-path tests covering:
  - `break outer;` from nested for-in loops
  - `break outer;` from nested while loops
  - `continue outer;` skipping outer iterations
  - Unlabeled break inside a labeled loop (unchanged behavior)
  - Named inner loop with `break inner;`
  - Labeled break inside function bodies (`compile_stmt_in_fn` path)
  - Labeled continue inside function bodies
- [x] All 4758 existing tests pass with zero regressions
- [x] `cargo fmt --all` clean

Closes #2484

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -7,6 +7,10 @@
     "resilient/src/vm.rs": {
       "branch": "res-2496-load-index-unchecked-panic",
       "claimed_at": "2026-05-25T12:30:00Z"
+    },
+    "resilient/src/compiler.rs": {
+      "branch": "res-2502-labeled-break-compile",
+      "claimed_at": "2026-05-25T09:24:24Z"
     }
   }
 }
--- a/resilient/src/compiler.rs
+++ b/resilient/src/compiler.rs
@@ -17,10 +17,10 @@ use std::collections::HashMap;
 
 /// Tracks break/continue patch sites accumulated while compiling a loop body.
 ///
-/// Created fresh for each While/ForIn loop. Nested loops create their own
-/// inner LoopState, shadowing the outer one, so `break`/`continue` always
-/// target the *innermost* enclosing loop — matching the tree-walker's
-/// `Value::Break`/`Value::Continue` bubble-up semantics.
+/// Created fresh for each While/ForIn loop. A `Vec<LoopState>` stack is
+/// threaded through the compilation so labeled `break`/`continue` can
+/// target an enclosing loop by name. Unlabeled break/continue always
+/// targets the *innermost* (last) entry.
 struct LoopState {
     /// PC of the back-edge target. For `while` this is the condition check;
     /// for `for-in` it's the index-increment code (set after body compilation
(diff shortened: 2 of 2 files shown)


EricSpencer00:

Violations:

EricSpencer00:
## Agent Handoff

- Time: `2026-05-25T09:32:22Z`
- Issue: #2484
- PR: #2485
- Branch: `res-2502-labeled-break-compile`
- Worktree: `unknown`
- Phase: `guardrail-red`

### Claimed files
- (none recorded)

### Resume summary

Violations:

PR #2512, opened 2026-05-25
RES-2528: add for-in loop support for maps in interpreter and VM

### LINKED ISSUE #2511: RES-2528: for-in loop over maps broken — interpreter rejects, VM silently produces wrong results
## Problem

`for x in map` does not work correctly in either execution mode:

**Interpreter**: `eval_for_in_in_scope` only handles `Value::Array` (line 24225). Iterating a map errors with "`for` iterable must be an array, got Map(...)".

**VM/Compiler**: `compile_for_in` computes `len(iterable)` then indexes with `iterable[0]`, `iterable[1]`, etc. via `LoadIndex`. For maps, `LoadIndex` treats the index as a key lookup (e.g., `map[Int(0)]`), which returns `Void` instead of the actual map entries. Additionally, the `len()` builtin doesn't support `Value::Map` — it errors at the length computation step.

This is a correctness bug for a safety-critical language — silent wrong results from `for x in map` in the VM.

## Solution

1. **`len()` builtin**: Add `Value::Map` support (return number of entries)
2. **Interpreter `eval_for_in`**: Handle `Value::Map` by iterating over keys (consistent with Python/JS semantics)
3. **VM**: Add `IterPrepare` opcode that normalizes iterables — arrays/strings pass through, maps are converted to their keys array. Emit this after evaluating the iterable in `compile_for_in`
4. **Tests**: Golden files and unit tests for `for k in map` in both interpreter and VM

## Acceptance criteria

- [ ] `for k in {"a" -> 1, "b" -> 2} { println(k); }` prints keys in both interpreter and VM
- [ ] `len({"a" -> 1})` returns 1
- [ ] `for k in map { map[k] }` works correctly to access values
- [ ] Existing for-in tests for arrays, strings, and ranges still pass

## Impact

Tier 2 correctness: maps are a core data structure and iterating them is a fundamental operation. Silent wrong results in the VM path are especially dangerous for safety-critical code.

---

EricSpencer00:
## Summary

- Maps are now iterable with `for k in map` — binds each key in deterministic sorted order
- Added `IterPrepare` VM opcode that normalizes iterables (maps → sorted keys array, arrays/strings pass through)
- `len()` builtin now supports maps
- 8 new tests covering interpreter and VM paths

## Problem

`for x in map` was broken in both execution modes:
- **Interpreter**: rejected maps with "iterable must be an array"
- **VM**: used `LoadIndex` with integer indices 0..len, which performed key lookups (`map[Int(0)]`) instead of positional iteration — producing `Void` for non-integer-keyed maps

## Design

The loop variable receives **keys** (matching Python/JS `for k in dict` semantics). Users access values via `map[k]` in the loop body. Keys are yielded in deterministic sorted order (Int < Str < Bool).

The VM fix adds a single `IterPrepare` opcode emitted after evaluating the for-in iterable. For maps, it extracts the sorted keys as an array; for arrays and strings, it's a no-op passthrough. The rest of the loop (`len` + sequential `LoadIndex`) works unchanged.

Closes #2511

## Test plan

- [x] `for k in {"b" -> 2, "a" -> 1}` yields keys `["a", "b"]` (interpreter)
- [x] `for k in {"x" -> 10, "y" -> 20} { total += m[k] }` sums to 30 (interpreter)
- [x] `len({"a" -> 1, "b" -> 2, "c" -> 3})` returns 3
- [x] VM: `for k in map` iterates keys in sorted order
- [x] VM: `for k in map { map[k] }` accesses values correctly
- [x] VM: empty map iteration executes zero body iterations
- [x] VM: integer-keyed maps iterate keys in numeric order
- [x] VM: `len(map)` returns entry count
- [x] All existing for-in tests for arrays, strings, ranges still pass

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,12 +1,20 @@
 {
   "claims": {
     "resilient/src/lib.rs": {
-      "branch": "res-2526-int-overflow-diagnostic",
-      "claimed_at": "2026-05-25T13:35:06Z"
+      "branch": "res-2528-map-iteration",
+      "claimed_at": "2026-05-25T13:53:15Z"
     },
-    "resilient/src/lexer_logos.rs": {
-      "branch": "res-2526-int-overflow-diagnostic",
-      "claimed_at": "2026-05-25T13:35:06Z"
+    "resilient/src/compiler.rs": {
+      "branch": "res-2528-map-iteration",
+      "claimed_at": "2026-05-25T13:53:15Z"
+    },
+    "resilient/src/vm.rs": {
+      "branch": "res-2528-map-iteration",
+      "claimed_at": "2026-05-25T13:53:15Z"
+    },
+    "resilient/src/bytecode.rs": {
+      "branch": "res-2528-map-iteration",
+      "claimed_at": "2026-05-25T13:53:15Z"
     }
   }
 }
--- a/resilient/src/bytecode.rs
+++ b/resilient/src/bytecode.rs
@@ -228,6 +228,13 @@ pub enum Op {
     ///
     /// Emitted for `expr?` (`Node::TryExpression`) nodes.
     TryUnwrap,
+    /// RES-2528: normalize an iterable for `for-in` loops. Pop TOS:
+    ///   - `Array` / `String` → push back unchanged (LoadIndex handles both).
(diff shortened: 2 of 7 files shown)


EricSpencer00:

PR #2518, opened 2026-05-25
RES-2534: sync handler-table dispatch with match-dispatch for LoadIndex/StoreIndex/Mul

### LINKED ISSUE #2517: RES-2534: handler-table dispatch missing Map support in LoadIndex/StoreIndex + string repeat size guard
## Problem
The handler-table dispatch path (`h_load_index`, `h_store_index`, `h_mul`) is out of sync with the match-dispatch path (`run_inner`). Three discrepancies:

1. **`h_load_index`** is missing Map support — `map[key]` fails under `Dispatch::Direct` but works under `Dispatch::Match`
2. **`h_store_index`** is missing Map support AND negative index wrapping — `map[key] = v` fails, and `arr[-1] = v` fails under Direct
3. **`h_mul`** (and the match-dispatch Mul path) are missing the `MAX_STRING_REPEAT` size guard that the tree-walker has — `"a" * 2_000_000_000` OOMs

## Solution
- Port Map check from match-dispatch `LoadIndex`/`StoreIndex` into `h_load_index`/`h_store_index`
- Add negative index wrapping to `h_store_index`
- Add string repeat size guard to both Mul dispatch paths

## Acceptance criteria
- [ ] `h_load_index` handles `Value::Map` with `MapKey::from_value`
- [ ] `h_store_index` handles `Value::Map` and negative index wrapping
- [ ] Both Mul paths reject repeat counts > 10M chars
- [ ] Test: map index read/write under `Dispatch::Direct`
- [ ] Test: negative array index under `Dispatch::Direct`
- [ ] Test: large string repeat returns error

## Impact
The two VM dispatch paths must produce identical results. These bugs silently produce wrong behavior for any user running with `RESILIENT_DISPATCH=direct`.

---

EricSpencer00:
## Summary

- **h_load_index**: Added Map support — the handler-table path only handled int-indexed containers (Array, Tuple, String), but the match-dispatch path also handles `Value::Map`. Map lookups via `m["key"]` would succeed under `Dispatch::Match` but crash or return wrong results under `Dispatch::Direct`.
- **h_store_index**: Added Map support AND negative array index wrapping — same Map gap as LoadIndex, plus `a[-1] = val` would panic or fail under Direct dispatch because negative indices weren't being resolved to positive offsets.
- **h_mul**: Added `MAX_STRING_REPEAT` (10M char) guard — `"x" * 100_000_000` would OOM under Direct dispatch while the match path already had the guard from the tree-walker.
- **op_to_index coverage test**: Added missing opcodes `IterPrepare`, `LoadGlobal`, `StoreGlobal`, `TryUnwrap`, `LoadIndexUnchecked` to the handler-table coverage smoke test.

## Test plan

- [x] `res2534_map_index_read_both_dispatch` — map read via `m["key"]` agrees on both dispatch paths
- [x] `res2534_map_index_write_both_dispatch` — map write via `m["key"] = val` agrees on both paths
- [x] `res2534_negative_array_index_both_dispatch` — `a[-1]` read agrees
- [x] `res2534_negative_array_store_both_dispatch` — `a[-1] = val` store agrees
- [x] `res2534_string_repeat_limit_both_dispatch` — huge repeat errors on both paths
- [x] `res2534_string_repeat_ok_both_dispatch` — normal repeat works on both paths

Closes #2517

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,3 +1,8 @@
 {
-  "claims": {}
+  "claims": {
+    "resilient/src/vm.rs": {
+      "branch": "res-2534-handler-table-sync",
+      "claimed_at": "2026-05-25T14:50:48Z"
+    }
+  }
 }
--- a/resilient/src/vm.rs
+++ b/resilient/src/vm.rs
@@ -340,6 +340,7 @@ fn err_at(line_info: &[u32], pc: usize, e: VmError) -> VmError {
 /// growth on pathologically-recursive input (test case for
 /// `VmError::CallStackOverflow`).
 const MAX_CALL_DEPTH: usize = 1024;
+const MAX_STRING_REPEAT: usize = 10_000_000;
 
 /// RES-081: one active function invocation. `chunk_idx = usize::MAX`
 /// marks the `main` frame; any other value indexes into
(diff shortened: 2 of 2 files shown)


EricSpencer00:

PR #2662, opened 2026-05-28
RES-2651: Option<T> type inference for pattern matching

### LINKED ISSUE #2651: Improve enum payload type inference for nested patterns (Some(Shape::Circle(r)))
## Summary

When matching on `Some(Shape::Circle(r))`, the inner `r` binding is typed as `Type::Any` because `Some(inner)` recurses with `scrut_ty = Type::Any` (the Option payload type is not tracked). This means type errors inside nested enum patterns are not caught by the static checker.

## Current behavior

```resilient
enum Shape { Circle(float) }
fn area(opt val) -> float {
    return match val {
        Some(Shape::Circle(r)) => r * r,  // r is typed as Any, no type checking
        _ => 0.0,
    };
}
```

The runtime works correctly, but incorrect code like `r + "string"` would pass type checking.

## Expected behavior

When the scrutinee of a `Some(...)` pattern is known to contain an enum value (e.g. from a parameter annotation or local variable type), the inner binding should carry the variant's declared payload type.

## Approach

This requires tracking element types through `Option<T>` — either by adding a type parameter to `Type::Option` or by propagating the scrutinee type through `Some` patterns when it is a known struct/enum type. The latter is simpler and can be done without generics.

## Scope

This is a follow-up to #2533 which fixed the base case of enum payload destructuring. Nested pattern type inference is a non-trivial addition that deserves its own PR.

Labels: `agent-ready`

EricSpencer00:
## Summary

Adds `Type::Option(Box<Type>)` to the type system so that `Some(expr)` constructions, `checked_*` builtins, and `Option<T>` annotations propagate the concrete inner type through pattern matching. Bindings inside `Some(x)` match arms now carry the real inner type instead of `Any`.

**Changes:**
- New `Type::Option(Box<Type>)` variant with Display, compatible(), and parse_type_name support
- `Some(expr)` call-site inference: `Some(42)` → `Option<int>`
- `checked_add/sub/mul/div` return `Option<int>` instead of `Any`
- `Pattern::Some(inner)` extracts inner type from scrutinee in `match_pattern_binding_types`
- Option exhaustiveness checking (both `Some` + `None` arms required)
- Argument type check uses `compatible()` instead of `!=` (fixes `Option<int>` vs `Option`)
- `unify.rs` extended with `Option` variant in apply/unify/occurs

## Test plan

- [x] New golden test `option_type_inference.rz` exercises all three inference paths (Some literal, checked_add, annotated parameter)
- [x] All 5,086+ existing tests pass
- [x] clippy clean, fmt clean

Closes #2651

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,16 +1,12 @@
 {
   "claims": {
     "resilient/src/lib.rs": {
-      "branch": "res-2660-static-assert",
-      "claimed_at": "2026-05-28T16:48:56Z"
+      "branch": "res-2651-nested-enum-type-inference",
+      "claimed_at": "2026-05-28T17:10:03Z"
     },
     "resilient/src/typechecker.rs": {
-      "branch": "res-2660-static-assert",
-      "claimed_at": "2026-05-28T16:48:56Z"
-    },
-    "resilient/src/lexer_logos.rs": {
-      "branch": "res-2660-static-assert",
-      "claimed_at": "2026-05-28T16:48:56Z"
+      "branch": "res-2651-nested-enum-type-inference",
+      "claimed_at": "2026-05-28T17:10:03Z"
     }
   }
 }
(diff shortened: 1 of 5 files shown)


EricSpencer00:

PR #2665, opened 2026-05-28
RES-2664: reject match arms with incompatible types

### LINKED ISSUE #2664: RES-2664: match arm type mismatch accepted silently — type safety hole
## Problem

The typechecker's `infer_common_arm_type` silently falls back to `Type::Any` when match arms return different types, accepting unsound programs without any diagnostic:

```resilient
fn test() -> int {
    let x = 1;
    return match x {
        1 => "hello",   // string
        _ => 42,         // int
    };
}
```

This compiles and runs, returning a string from a function declared `-> int`. The root cause is line 250 of `typechecker.rs`: `infer_common_arm_type` returns `Type::Any` on disagreement instead of reporting an error.

## Solution

Change `infer_common_arm_type` to return `Result<Type, String>` and propagate a type mismatch error when arms disagree. The error message should show the conflicting types and their arm locations.

## Acceptance criteria

- [ ] Match with arms of different types (`string` vs `int`) produces a type error
- [ ] Match with all arms of the same type continues to work
- [ ] Match with `Any` arms (untyped expressions) continues to work
- [ ] Golden test proving the error is emitted
- [ ] No false positives on existing test suite

## Impact

**Critical for a safety-critical compiler.** Match expressions are the primary control-flow mechanism for pattern matching in Resilient. Silently accepting heterogeneous arm types means the type system cannot guarantee value-level type safety through match — a fundamental soundness violation.

---

EricSpencer00:
## Summary

Fixes a **type safety hole** where match expressions with arms returning different types (e.g. `string` vs `int`) were silently accepted. The typechecker's `infer_common_arm_type` fell back to `Type::Any` on disagreement instead of raising an error.

**Before:** `match x { 1 => "hello", _ => 42 }` compiled silently, even inside `fn foo() -> int`.

**After:** `match arms have incompatible types: string and int` error raised.

**Changes:**
- Add explicit compatibility check between concrete arm types in the match expression handler
- Uses `compatible()` for structural comparison (handles Option<T>, type aliases, etc.)
- Three new unit tests in `res402_arm_type_inference` module
- Golden test `match_arm_type_check.rz` with valid same-type match patterns

## Test plan

- [x] New test `match_arms_type_mismatch_errors` — string vs int in match arms → error
- [x] New test `match_arms_three_way_mismatch_errors` — string vs bool vs int → error
- [x] New test `match_arms_same_type_accepted` — all-int arms → no error
- [x] Golden test with Option, wildcard, and multiple patterns
- [x] All 5,089 existing tests pass (no false positives)
- [x] clippy clean, fmt clean

Closes #2664

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,16 +1,12 @@
 {
   "claims": {
     "resilient/src/lib.rs": {
-      "branch": "res-2632-struct-update-syntax",
-      "claimed_at": "2026-05-28T17:42:07Z"
+      "branch": "res-2664-match-arm-type-mismatch",
+      "claimed_at": "2026-05-28T18:16:17Z"
     },
     "resilient/src/typechecker.rs": {
-      "branch": "res-2632-struct-update-syntax",
-      "claimed_at": "2026-05-28T17:42:07Z"
-    },
-    "resilient/src/lexer_logos.rs": {
-      "branch": "res-2632-struct-update-syntax",
-      "claimed_at": "2026-05-28T17:42:07Z"
+      "branch": "res-2664-match-arm-type-mismatch",
+      "claimed_at": "2026-05-28T18:16:17Z"
     }
   }
 }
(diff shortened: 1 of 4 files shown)


EricSpencer00:

PR #2667, opened 2026-05-29
RES-2666: fix compiler panic on i64::MIN/MAX literals

### LINKED ISSUE #2666: RES-2666: mutation-testing off-by-one mutant generation panics on i64::MIN/MAX literals
## Problem

The mutation-testing analysis pass (`resilient/src/mutation_testing.rs`)
generates "off-by-one" mutant descriptions for every integer literal by
computing `value - 1` and `value + 1` on the raw `i64` literal value:

```rust
Node::IntegerLiteral { value, .. } => {
    ...
    push(out, fn_name, "literal", format!("off-by-one `{value}` -> `{}`", value - 1));
    push(out, fn_name, "literal", format!("off-by-one `{value}` -> `{}`", value + 1));
}
```

When the source program contains `i64::MAX` (`9223372036854775807`),
`value + 1` overflows and **panics the compiler**. When it contains
`i64::MIN`, `value - 1` overflows and panics. Because mutation analysis
runs as part of resilience scoring on `rz run`, `rz check`, and
`rz test`, any program containing an extremal integer literal crashes
the whole compiler instead of producing output.

Reproduction:

```
fn main() {
    let a = 9223372036854775807;
    println(a + 1);
}
main();
```

```
thread '<unnamed>' panicked at resilient/src/mutation_testing.rs:159:61:
attempt to add with overflow
```

A panic in the compiler on valid input is a bug (CLAUDE.md: "A panic in
the compiler is a bug"). This is especially bad for a safety-critical
language toolchain.

## Solution

Use `checked_add` / `checked_sub` for the off-by-one mutant arithmetic
and only emit the mutant when it does not overflow. The const-fold pass
already follows this discipline (`checked_*` everywhere); mutation
testing should too. For `i64::MAX` only the `-1` mutant is emitted; for
`i64::MIN` only the `+1` mutant. No off-by-one mutant is lost for any
non-extremal literal.

## Acceptance criteria
- [ ] Compiling/running a program containing `i64::MAX` or `i64::MIN`
      literals no longer panics.
- [ ] A regression test asserts mutation generation does not panic on
      extremal literals and still emits the in-range off-by-one mutant.
- [ ] Off-by-one mutants for ordinary literals are unchanged.

## Impact

Removes a compiler-panic crash class triggered by ordinary source
programs. Hardens the resilience-scoring path that runs on every
`rz run`/`check`/`test`.

---

EricSpencer00:
Closes #2666

## Problem

The mutation-testing analysis pass generated "off-by-one" mutant
descriptions for every integer literal with unchecked arithmetic:

```rust
format!("off-by-one `{value}` -> `{}`", value - 1)
format!("off-by-one `{value}` -> `{}`", value + 1)
```

For `i64::MAX` literals, `value + 1` overflows and panics the compiler;
for `i64::MIN`, `value - 1` overflows. Because resilience scoring runs
this pass on every `rz run`/`rz check`/`rz test`, any program containing
an extremal integer literal crashed the toolchain:

```
fn main() { let a = 9223372036854775807; println(a + 1); }
```
```
thread '<unnamed>' panicked at resilient/src/mutation_testing.rs:159:61:
attempt to add with overflow
```

## Fix

Use `checked_sub`/`checked_add` and skip the direction that would
overflow. `i64::MAX` keeps its `-1` mutant, `i64::MIN` keeps its `+1`
mutant, and every non-extremal literal is unchanged. This mirrors the
discipline already used by the const-fold pass.

## Tests

- `extremal_int_literals_do_not_panic` — an `i64::MAX` literal generates
  mutants without panicking and still emits its in-range `-1` mutant.
- `ordinary_literal_off_by_one_unchanged` — both off-by-one mutants are
  still emitted for ordinary literals.

Full suite: 2167 passed, 0 failed. fmt + clippy clean.

### DIFF
--- a/resilient/src/mutation_testing.rs
+++ b/resilient/src/mutation_testing.rs
@@ -146,18 +146,26 @@ fn generate_in<'a>(node: &'a Node, fn_name: &'a str, out: &mut Vec<Mutation<'a>>
                 push(out, fn_name, "literal", "swap `0` -> `1`");
             } else {
                 push(out, fn_name, "literal", format!("swap `{value}` -> `0`"));
-                push(
-                    out,
-                    fn_name,
-                    "literal",
-                    format!("off-by-one `{value}` -> `{}`", value - 1),
-                );
-                push(
-                    out,
-                    fn_name,
-                    "literal",
-                    format!("off-by-one `{value}` -> `{}`", value + 1),
-                );
+                // Off-by-one mutants use checked arithmetic: `i64::MIN`
+                // has no `-1` mutant and `i64::MAX` has no `+1` mutant,
+                // so skip the direction that would overflow rather than
+                // panicking the compiler on extremal literals.
+                if let Some(dec) = value.checked_sub(1) {
+                    push(
+                        out,
+                        fn_name,
+                        "literal",
+                        format!("off-by-one `{value}` -> `{dec}`"),
+                    );
+                }
+                if let Some(inc) = value.checked_add(1) {
+                    push(
+                        out,
+                        fn_name,
+                        "literal",
+                        format!("off-by-one `{value}` -> `{inc}`"),
+                    );
+                }
             }
         }
         Node::BooleanLiteral { value, .. } => {


EricSpencer00:

EricSpencer00:

PR #2692, opened 2026-05-30
RES-2691: float literals assignable to f32-annotated variables

### LINKED ISSUE #2691: RES-2691: float literals not assignable to f32-annotated variables
## Problem

\`let x: f32 = 3.14\` produces a type error: _\"let x: f32 — value has type float\"_.

Float literals infer as \`Type::Float\` (64-bit), but the \`compatible()\` function in the typechecker has no rule allowing \`Type::Float\` to satisfy a \`Type::Float32\` annotation. The identical pattern for integers is already handled: integer literals (\`Type::Int\`) are accepted wherever a pinned-int type (Int8/Int16/Int32/etc.) is expected.

## Solution

Add two symmetric cases to \`compatible()\` in \`typechecker.rs\` mirroring the Int ↔ pinned-int rules:

\`\`\`rust
// Float literals (Type::Float) → f32-annotated variable (Type::Float32)
// and vice-versa — same pattern as Int ↔ pinned-int.
if (*a == Type::Float32 && *b == Type::Float) || (*a == Type::Float && *b == Type::Float32) {
    return true;
}
\`\`\`

Note: \`unify()\` (used for arithmetic) already errors on Float/Float32 mixing to prevent implicit cross-width arithmetic. This change only affects assignment/annotation checking.

## Acceptance criteria
- [ ] \`let x: f32 = 3.14\` compiles without type error
- [ ] \`fn foo() -> f32 { return 1.0; }\` compiles without type error
- [ ] \`3.14 + as_f32(1.0)\` (arithmetic) still errors with the cross-width message
- [ ] Golden sidecar added in \`resilient/examples/\`

## Impact

Users cannot currently use f32 type annotations with literal values — every float program that declares \`f32\` variables needs a redundant \`as_f32()\` cast even for plain literals. Fixes a silent usability gap introduced when Float32 was added.

---

EricSpencer00:
## Summary

Float literals infer as `Type::Float` (f64), but `compatible()` in the typechecker had no rule allowing them to satisfy a `Type::Float32` annotation. This made `let x: f32 = 3.14` produce a spurious type error even though f32 is semantically a float subtype.

**Root cause:** The `compatible()` function handles `Int` ↔ pinned-int (Int8, Int16, etc.) coercions but was missing the symmetric `Float` ↔ `Float32` case.

**Fix:** Two symmetric cases added to `compatible()` (6 lines), exactly mirroring the Int ↔ pinned-int pattern. The `unify()` path (arithmetic) already errors on Float/Float32 mixing and is unchanged.

## Changes

- `resilient/src/typechecker.rs`: `compatible()` — add Float ↔ Float32 rule
- `resilient/src/float32.rs`: three new tests (annotation accepted, arithmetic accepted, cross-width arithmetic still errors)
- `resilient/examples/f32_annotation.rz` + `f32_annotation.expected.txt`: end-to-end golden

## Test changes

Added 3 tests to `float32::tests`:
1. `f32_let_annotation_accepts_float_literal` — verifies `let x: f32 = 3.14` works
2. `f32_let_annotation_two_vars_arithmetic` — verifies annotated-f32 arithmetic works
3. `f32_cross_width_arithmetic_still_errors` — verifies `float_var + f32_var` still type-errors (calls `TypeChecker::new().check_program()` directly since `run_program()` skips the typechecker)

Closes #2691

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,8 +1,8 @@
 {
   "claims": {
-    "resilient/src/compiler.rs": {
-      "branch": "res-2689-vm-blanket-impl-noop",
-      "claimed_at": "2026-05-30T09:24:53Z"
+    "resilient/src/typechecker.rs": {
+      "branch": "res-2691-f32-annotation-compat",
+      "claimed_at": "2026-05-30T09:34:11Z"
     }
   }
 }
--- a/resilient/examples/f32_annotation.expected.txt
+++ b/resilient/examples/f32_annotation.expected.txt
@@ -1,0 +1,5 @@
+3.14
+2
+5.140000000000001
+5.140000000000001
+Program executed successfully
--- a/resilient/examples/f32_annotation.rz
+++ b/resilient/examples/f32_annotation.rz
@@ -1,0 +1,24 @@
+// RES-2691: float literals are directly assignable to f32-annotated variables.
+// Prior to this fix, `let x: f32 = 3.14` produced a type error.
+
+fn add(f32 a, f32 b) -> f32 {
+    return a + b;
+}
+
+fn main() -> void {
+    // Direct float literal → f32 annotation (was a type error before RES-2691).
+    let x: f32 = 3.14;
+    let y: f32 = 2.0;
+    println(x);
+    println(y);
+
+    // f32 arithmetic with annotated variables.
+    let z: f32 = x + y;
+    println(z);
+
+    // f32 return value from a function with f32 parameters.
+    let w: f32 = add(x, y);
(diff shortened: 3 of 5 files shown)


EricSpencer00:

EricSpencer00:

EricSpencer00:
## Agent Handoff

- Time: `2026-05-30T09:46:46Z`
- Issue: #2691
- PR: #2692
- Branch: `res-2691-f32-annotation-compat`
- Worktree: `unknown`
- Phase: `guardrail-green`

### Claimed files
- (none recorded)

### Resume summary

PR #2698, opened 2026-05-30
RES-2697: implement default trait method bodies

### LINKED ISSUE #2697: RES-2697: default trait method bodies — parse and dispatch correctly
## Problem

Trait methods with default bodies parse without error but don't work. Calling a default method on a struct that doesn't override it fails at runtime:

```rz
trait Greet {
    fn greet(self) -> string;
    fn greet_loudly(self) -> string {  // default body
        return "HELLO!";
    }
}
struct Alice { int x }
impl Greet for Alice {
    fn greet(self) -> string { return "hi"; }
    // greet_loudly not overridden — should use default
}
let a = new Alice { x: 0 };
println(a.greet_loudly());  // Runtime error: Struct Alice has no field 'greet_loudly'
```

The parser comment says: "the body is allowed but ignored — keeps the parser tolerant of stub bodies that future tickets may grow into default methods." That future ticket is this one.

## Solution

1. Add `default_body: Option<Box<Node>>` to `TraitMethodSig` in `traits.rs`
2. Parse the body (not skip it) in `parse_method_sig` by calling the main parser's function
3. When evaluating `Node::TraitDecl` in the interpreter, register default methods as `TraitName$default$method_name` values
4. When evaluating `Node::ImplBlock { trait_name: Some(t), ... }`, for each trait method not overridden but having a default body, register `StructName$method_name` from the default
5. In `traits::check`, don't require a missing method if it has a default body

## Acceptance criteria
- [ ] Default method on trait can be called without override
- [ ] Override takes priority over default
- [ ] Trait check passes impl that only provides non-default methods
- [ ] Golden test + unit tests

---

EricSpencer00:
## Problem
Trait declarations can include method bodies as defaults (e.g. `fn greet_loudly(self) { ... }`), but the interpreter discards them at parse time. Calling a default method on a struct that doesn't override it fails at runtime with `Struct X has no field 'method_name'`.

## Solution
1. **Parse**: extend `TraitMethodSig` with `params: Vec<String>` and `default_body: Option<Box<Node>>`. When `parse_method_sig` encounters `{`, call `parser.parse_block_statement()` to capture the full body node.
2. **Typecheck**: in `traits::check`, allow a missing impl method if the trait declares a default body for it.
3. **Runtime**: add `trait_method_defaults: Rc<RefCell<HashMap<...>>>` to `Interpreter`. When evaluating a `TraitDecl`, register each method's default body as a `Value::Function`. When evaluating an `ImplBlock`, inject any un-overridden defaults as `StructName$method_name` entries so normal method dispatch finds them.

## Test plan
- Golden example `default_method.rz` / `.expected.txt` calling an un-overridden default method
- Unit test: default fires when not overridden; overriding replaces it; call still works

Closes #2697

🤖 Generated with Claude Code

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,8 +1,12 @@
 {
   "claims": {
     "resilient/src/traits.rs": {
-      "branch": "res-2695-where-projection-checking",
-      "claimed_at": "2026-05-30T10:16:44Z"
+      "branch": "res-2697-default-trait-methods",
+      "claimed_at": "2026-05-30T10:40:04Z"
+    },
+    "resilient/src/lib.rs": {
+      "branch": "res-2697-default-trait-methods",
+      "claimed_at": "2026-05-30T10:40:04Z"
     }
   }
 }
--- a/resilient/examples/default_method.expected.txt
+++ b/resilient/examples/default_method.expected.txt
@@ -1,0 +1,5 @@
+Hello, Alice
+Hello, Alice!!!
+Hi, Bob
+HEY, Bob!
+Program executed successfully
--- a/resilient/examples/default_method.rz
+++ b/resilient/examples/default_method.rz
@@ -1,0 +1,41 @@
+// RES-2697: default trait method bodies.
+// `greet_loudly` is not overridden by Alice's impl — the trait default fires.
+// Bob overrides it, so Bob's version runs instead.
+
+trait Greet {
+    fn greet(self) -> string;
+
+    fn greet_loudly(self) -> string {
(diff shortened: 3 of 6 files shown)


EricSpencer00:

EricSpencer00:
## Agent Handoff

- Time: `2026-05-30T10:59:55Z`
- Issue: #2697
- PR: #2698
- Branch: `res-2697-default-trait-methods`
- Worktree: `unknown`
- Phase: `guardrail-green`

### Claimed files
- (none recorded)

### Resume summary

PR #2739, opened 2026-05-30
RES-2738: String dot-call methods — replace, chars, reverse, strip_prefix, strip_suffix, lines

### LINKED ISSUE #2738: RES-2738: String dot-call methods — replace, chars, reverse, strip_prefix, strip_suffix, lines
## Problem

Several string methods are registered in the typechecker as valid dot-call methods but missing from the runtime String method dispatch allowlist:

- `s.replace(from, to)` → "Cannot access field 'replace' on non-struct String"
- `s.chars()` → "Cannot access field 'chars' on non-struct String"

Additionally, the typechecker has a pre-existing arity bug: `chars` is grouped with `split` (1 param), but `chars()` takes 0 extra args (the receiver is the only arg). This causes a false type error.

Other commonly-needed string operations (`reverse`, `strip_prefix`, `strip_suffix`, `lines`) are available as `string_reverse`, `string_strip_prefix`, etc. but have no dot-call form.

## Solution

1. Refactor the combined String/Array runtime dispatch block into two separate type-specific blocks. The String block uses an explicit short-name → full-builtin-name mapping (like the RES-2736 Map/Set blocks) so names like `reverse` can route to `string_reverse` without conflicting with the Array `reverse`.

2. Add these methods to the String dispatch:
   - `replace(from, to)` → `replace`
   - `chars()` → `string_chars`
   - `reverse()` → `string_reverse`
   - `strip_prefix(prefix)` → `string_strip_prefix`
   - `strip_suffix(suffix)` → `string_strip_suffix`
   - `lines()` → `string_lines`

3. Fix the typechecker: separate `chars` from `split` so `chars` has `params: vec![]` (0 extra args). Add type registrations for `reverse`, `strip_prefix`, `strip_suffix`, `lines`.

## Acceptance criteria
- [ ] `s.replace("old", "new")` works as a method call
- [ ] `s.chars()` returns array of char values without type error
- [ ] `s.reverse()` returns the reversed string
- [ ] `s.strip_prefix("x")` returns Ok/Err
- [ ] `s.strip_suffix("x")` returns Ok/Err
- [ ] `s.lines()` splits by newline
- [ ] Existing string methods (trim, split, contains, etc.) still work
- [ ] No false arity errors from typechecker
- [ ] Golden example + unit tests

## Impact

`replace` is one of the most common string operations in any language. It was registered in the typechecker but not the runtime — users who try `s.replace(...)` get a confusing runtime error.

---

EricSpencer00:
## Problem

Several string methods were registered in the typechecker but absent from the runtime dispatch allowlist:
- `s.replace("old", "new")` → \"Cannot access field 'replace' on non-struct String\"
- `s.chars()` → \"Cannot access field 'chars' on non-struct String\"

The typechecker also had a pre-existing arity bug: `chars` was grouped with `split` (1-param), causing a false \"Expected 1 arguments, got 0\" error even though `chars()` takes no extra args.

## Changes

**`resilient/src/lib.rs`**
- Refactored the combined `String | Array` method dispatch into two separate type-specific blocks
- String block now uses an explicit `short_name → full_builtin_name` map (like RES-2736 Map/Set), so `reverse` can route to `string_reverse` without colliding with Array's `reverse` → `array_reverse`
- New String methods: `replace`, `chars`, `reverse`, `strip_prefix`, `strip_suffix`, `lines`
- 8 unit tests in `res2738_string_dot_methods`

**`resilient/src/typechecker.rs`**
- Fixed: `chars` separated from `split` — now correctly has `params: vec![]` (0 extra args)
- Added type registrations: `chars`, `lines` (→ Array), `reverse` (→ String), `strip_prefix`, `strip_suffix` (→ Result)

**`resilient/examples/string_dot_methods.rz` + `.expected.txt`**  
- Golden example exercising all new string methods

## Test changes

Added `res2738_string_dot_methods` tests including `string_chars_method_no_type_error` which specifically verifies the typechecker arity fix (no false error on `s.chars()`).

## Before / After

```
// Before:
let r = replace(s, "old", "new");       // global function only
let cs = string_chars(s);              // global function only
// s.replace(...) and s.chars() both failed at runtime

// After:
let r = s.replace("old", "new");
let cs = s.chars();
let rev = s.reverse();
let lines = s.lines();
```

Closes #2738

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,8 +1,12 @@
 {
   "claims": {
     "resilient/src/lib.rs": {
-      "branch": "res-2736-map-set-dot-methods",
-      "claimed_at": "2026-05-30T16:06:46Z"
+      "branch": "res-2738-string-method-dispatch",
+      "claimed_at": "2026-05-30T16:20:34Z"
+    },
+    "resilient/src/typechecker.rs": {
+      "branch": "res-2738-string-method-dispatch",
+      "claimed_at": "2026-05-30T16:20:34Z"
     }
   }
 }
--- a/resilient/examples/string_dot_methods.expected.txt
+++ b/resilient/examples/string_dot_methods.expected.txt
@@ -1,0 +1,11 @@
+Hello, Resilient!
+[H, e, l, l, o, ,,  , W, o, r, l, d, !]
+!dlroW ,olleH
+["Hello, World!"]
+hello
+value
+prefix_value
+value
+a | b | c
+["one", "two", "three"]
+Program executed successfully
--- a/resilient/examples/string_dot_methods.rz
+++ b/resilient/examples/string_dot_methods.rz
@@ -1,0 +1,26 @@
+// RES-2738: String dot-call methods — replace, chars, reverse,
+// strip_prefix, strip_suffix, lines. Previously available only as
+// global functions (replace/string_*) or not at all via dot notation.
+
+let s = "Hello, World!";
+println(s.replace("World", "Resilient"));
+println(s.chars());
+println(s.reverse());
+println(s.lines());
+
+let padded = "  hello  ";
+println(padded.trim());
+
+let prefixed = "prefix_value";
+println(prefixed.strip_prefix("prefix_"));
+println(prefixed.strip_prefix("other_"));
+
+let suffixed = "value.txt";
+println(suffixed.strip_suffix(".txt"));
+
+let csv = "a,b,c";
+let parts = csv.split(",");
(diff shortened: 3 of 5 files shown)


EricSpencer00:

EricSpencer00:
## Agent Handoff

- Time: `2026-05-30T16:29:47Z`
- Issue: #2738
- PR: #2739
- Branch: `res-2738-string-method-dispatch`
- Worktree: `unknown`
- Phase: `guardrail-green`

### Claimed files
- (none recorded)

### Resume summary

PR #2742, opened 2026-05-30
RES-2740/RES-2741: fix Result<T,E> annotation false-positive and struct update trailing spread

### LINKED ISSUE #2740: RES-2740: Result<T, E> return-type annotation always produces false-positive type error
## Problem
Any function annotated with a parametric result type triggers a false-positive 'return type mismatch':
\`\`\`
fn f() -> Result<int, string> { return Ok(42); }
// → Type error: return type mismatch — declared Result<int, string>, returning Result
\`\`\`

Root cause: \`parse_type_name_inner\` in \`typechecker.rs\` handles bare \`"Result"\` → \`Type::Result\` and \`"Option<T>"\` → \`Type::Option(Box<Type>)\`, but has no case for \`"Result<T, E>"\`. The string \`"Result<int, string>"\` falls through to the alias lookup and then to \`Ok(Type::Struct("Result<int, string>"))\`. The returned \`Type::Result\` from \`Ok()\`/\`Err()\` does not match \`Type::Struct("Result<int, string>")\`, causing the false error.

## Solution
Add a \`Result<...>\` pattern arm immediately after the \`Option<T>\` arm in \`parse_type_name_inner\`:
\`\`\`rust
other if other.starts_with("Result<") && other.ends_with('>') => Ok(Type::Result),
\`\`\`
Since \`Type::Result\` is currently unparameterized at MVP level, we just discard the type params and resolve as plain \`Type::Result\`.

## Acceptance criteria
- [ ] \`fn f() -> Result<int, string> { return Ok(42); }\` compiles without error
- [ ] \`fn f() -> Result<int, string> { return Err("bad"); }\` compiles without error  
- [ ] Existing Result tests still pass
- [ ] Golden file added

## Impact
This bug makes ALL idiomatic error-handling functions with explicit return types fail. It's the highest-friction bug for new users.

---

### LINKED ISSUE #2741: RES-2741: struct update `..base` at end of field list fails — only works at start
## Problem
Struct update syntax only works when \`..base\` appears at the START of the field list:
\`\`\`
// Works:
new Point { ..base, x: 10 }

// Fails with parser error:
new Point { x: 10, ..base }
// → Expected field name in struct literal, found \`..'\`
\`\`\`

Root cause: The parser's struct-literal field loop matches only \`Token::Identifier\` at the start of each field. When it encounters \`Token::DotDot\` (i.e. \`..\`), it records an error and breaks. The \`..base\` case is only checked before the loop begins.

Rust and most languages allow \`..base\` at the END of a struct literal (after explicit field assignments), which is the more common usage pattern.

## Solution
In the struct-literal field loop (around line 9157 in \`lib.rs\`), add a check for \`Token::DotDot\` at the top of each iteration. If encountered, parse the base expression, set \`base\`, break out of the loop (trailing spread must be last).

## Acceptance criteria
- [ ] \`new Point { x: 10, ..base }\` parses and evaluates correctly
- [ ] \`new Point { ..base, x: 10 }\` still works (start position)
- [ ] \`new Point { x: 10, y: 20, ..base }\` works (multiple fields before spread)
- [ ] Golden file added

## Impact
Users following Rust conventions write \`..base\` at the end. The current position restriction is undocumented and counterintuitive.

---

EricSpencer00:
## Summary

Two P0 typechecker/parser correctness fixes:

### RES-2740: `Result<T, E>` parametric return-type annotation produces false-positive type error

**Root cause:** `parse_type_name_inner` had an arm for bare `"Result"` → `Type::Result` and for `"Option<T>"` → `Type::Option(...)`, but no arm for `"Result<T, E>"`. The string `"Result<int, string>"` fell through to the alias lookup and then to `Type::Struct("Result<int, string>")`, which never compared equal to `Type::Result` from `Ok()`/`Err()` checks.

**Fix:** Added `other if other.starts_with("Result<") && other.ends_with('>')` → `Ok(Type::Result)` immediately after the `Option<T>` arm. Since `Type::Result` is unparameterized at MVP level the params are discarded.

### RES-2741: Struct update `..base` only works at the START of a field list

**Root cause:** The struct-literal field loop opened with `match current_token { Token::Identifier(n) => ..., _ => error }`. When a user wrote `new P { x: 10, ..base }`, the parser got to `..` after the comma and hit the wildcard arm, recording an error.

**Fix:** Added a `Token::DotDot` check at the top of each loop iteration. If encountered, parse the base expression, set `base`, and break — exactly mirroring the start-of-list handling already present.

## Test plan
- [x] `fn f() -> Result<int, string> { return Ok(42); }` compiles without error
- [x] `fn f() -> Result<int, string> { return Err("bad"); }` compiles without error
- [x] `new P { x: 5, ..base }` parses and evaluates correctly
- [x] `new P { x: 1, y: 2, ..base }` parses and evaluates correctly
- [x] `new P { ..base, z: 99 }` (leading spread) still works
- [x] All existing tests pass

Closes #2740
Closes #2741

### DIFF
--- a/agent-scripts/file-claims.json
+++ b/agent-scripts/file-claims.json
@@ -1,12 +1,12 @@
 {
   "claims": {
     "resilient/src/lib.rs": {
-      "branch": "res-2738-string-method-dispatch",
-      "claimed_at": "2026-05-30T16:20:34Z"
+      "branch": "res-2740-result-struct-update-fixes",
+      "claimed_at": "2026-05-30T16:38:34Z"
     },
     "resilient/src/typechecker.rs": {
-      "branch": "res-2738-string-method-dispatch",
-      "claimed_at": "2026-05-30T16:20:34Z"
+      "branch": "res-2740-result-struct-update-fixes",
+      "claimed_at": "2026-05-30T16:38:34Z"
     }
   }
 }
--- a/resilient/examples/result_parametric_annotation.expected.txt
+++ b/resilient/examples/result_parametric_annotation.expected.txt
@@ -1,0 +1,5 @@
+5
+division by zero
+42
+not positive
+Program executed successfully
--- a/resilient/examples/result_parametric_annotation.rz
+++ b/resilient/examples/result_parametric_annotation.rz
@@ -1,0 +1,41 @@
+// RES-2740: Result<T, E> parametric return-type annotations — previously
+// produced a false-positive "return type mismatch" error. The annotation
+// is now accepted and type-checked as plain Result (params not tracked at MVP).
+
+fn divide(int a, int b) -> Result<int, string> {
+    if b == 0 {
+        return Err("division by zero");
+    }
+    return Ok(a / b);
+}
+
+fn to_positive(int n) -> Result<int, string> {
+    if n < 0 {
+        return Err("not positive");
+    }
+    return Ok(n);
+}
+
+let r1 = divide(10, 2);
+match r1 {
+    Ok(v) => println(v),
+    Err(e) => println(e)
+}
+
+let r2 = divide(10, 0);
+match r2 {
+    Ok(v) => println(v),
+    Err(e) => println(e)
+}
+
+let r3 = to_positive(42);
+match r3 {
+    Ok(v) => println(v),
+    Err(e) => println(e)
+}
+
+let r4 = to_positive(-5);
+match r4 {
+    Ok(v) => println(v),
+    Err(e) => println(e)
+}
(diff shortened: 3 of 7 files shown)


EricSpencer00:

EricSpencer00:
## Agent Handoff

- Time: `2026-05-30T16:54:45Z`
- Issue: #2740
- PR: #2742
- Branch: `res-2740-result-struct-update-fixes`
- Worktree: `unknown`
- Phase: `guardrail-green`

### Claimed files
- (none recorded)

### Resume summary

PR #2815, opened 2026-05-31
RES-2814: fix false-positive type errors on generic enum constructors

### LINKED ISSUE #2814: RES-2814: fix false-positive type errors on generic enum constructors
## Problem

The typechecker rejects valid generic enum constructor calls like:
```
enum Either<L, R> { Left(L), Right(R) }
let a = Either::Left(42)  // Error: argument has type int, expected L
```

The error occurs because `parse_type_name("L")` produces `Type::Struct("L")` and
`compatible(Struct("L"), Int)` returns false. The typechecker doesn't know that `L` is
a type parameter, not a concrete type.

Additionally, match-arm destructuring binds extracted values to `Type::Struct("L")` instead
of `Type::Any`, causing downstream "Cannot compare L and int" errors.

## Solution

1. Store enum generic type parameters in `enum_type_params: HashMap<String, Vec<String>>`
2. During constructor validation, skip the type compatibility check when the declared param
   type is a generic type parameter — accept any concrete argument
3. During match-arm payload binding resolution, resolve generic type parameter names to
   `Type::Any` instead of `Type::Struct("L")`

## Acceptance criteria

- [x] `generic_enums.rz` golden test passes without type errors
- [x] Non-generic enum constructors still validate argument types
- [x] Constructor arity is still enforced for generic enums
- [x] Match-arm destructured bindings are usable (no "Cannot compare L" errors)

## Impact

Tier 1 compiler correctness: fixes a false-positive type error that blocks legitimate
generic enum usage. Generic enums (`Either`, `Box`, `Tree`) are a core language feature
(RES-2575) that the typechecker was rejecting.

---

EricSpencer00:
## Summary

Fixes the typechecker rejecting valid generic enum constructor calls.

**Before:** `Either::Left(42)` → Type error: argument has type int, expected L
**After:** Accepted — `L` is recognized as a type parameter and unified with any concrete type.

### Changes

- Store enum type parameters in `enum_type_params` field alongside `enum_decls`
- Constructor validation: skip compatibility check when declared param type is a generic type parameter
- Match-arm payload binding: resolve generic type param names to `Type::Any` so destructured values are usable
- 5 new unit tests covering generic constructors, match bindings, arity enforcement, and non-generic validation

### Impact

Tier 1 compiler correctness fix. Fixes `generic_enums.rz` golden test — generic enums (RES-2575) now type-check correctly.

Closes #2814

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/resilient/src/typechecker.rs
+++ b/resilient/src/typechecker.rs
@@ -1403,6 +1403,11 @@ pub struct TypeChecker {
     /// payload, all duplicated per lookup). Mirrors RES-1363's
     /// `contract_table` and RES-1365's `struct_fields` refactors.
     pub(crate) enum_decls: HashMap<String, std::sync::Arc<Vec<crate::EnumVariant>>>,
+    /// RES-2814: type parameters declared on generic enums. Populated
+    /// alongside `enum_decls` so the constructor validator can treat
+    /// declared type-param names (e.g. `L`, `R` in `Either<L, R>`)
+    /// as wildcards rather than concrete struct types.
+    enum_type_params: HashMap<String, Vec<String>>,
     /// RES-128: alias name → raw target type name. Populated by
     /// every `TypeAlias` node. Consulted by `parse_type_name` which
     /// walks the chain (with a `seen` set for cycle detection) and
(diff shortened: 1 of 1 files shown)


EricSpencer00:

Issue #2828, opened 2026-06-01
Continuous / hybrid-system dynamics: design doc for flow semantics and solver boundary


EricSpencer00:
## Goal

Write the design doc that makes hybrid-system work implementable without guessing later.

## Scope

Produce a document that locks down the surface syntax and verification boundary for continuous modes and hybrid transitions.

## Suggested files / modules

- `docs/hybrid_dynamics_design.md`
- `docs/AI_THREAT_MODEL.md` only if cross-linking is useful for style/reference
- Existing time / verifier docs that explain the `now` model

## Acceptance criteria

- [ ] New doc defines the proposed syntax for flow declarations, guards, and resets.
- [ ] Doc explains the time model decision and how it interacts with the existing `now` idiom.
- [ ] Doc separates what is checked in-tree from what is delegated to external tools such as dReal or Flow*.
- [ ] Doc includes at least two worked examples and a short list of unresolved follow-on questions.

## Verification

- [ ] The doc stands on its own for a future implementation ticket and references the next likely code touchpoints.

## Out of scope

- Parser or verifier code
- Solver integration
- Runtime semantics changes outside the design document

Issue #2827, opened 2026-06-01
Probabilistic property bounds: loop-free closed-form checker MVP


EricSpencer00:
## Goal

Land the first executable probabilistic-checking slice by validating simple probability bounds over loop-free weighted choice blocks.

## Scope

- Accept and check `#[prob(...)]` annotations against explicit weighted branches.
- Support only finite, loop-free probabilistic blocks in this ticket.
- Reject unsupported loops / fixpoint cases with a clear diagnostic instead of silently accepting them.

## Suggested files / modules

- `resilient/src/probabilistic_contracts.rs`
- Parser / attribute handling where `#[prob(...)]` is recorded
- Typechecker or verifier path that evaluates the bound
- Focused tests under `resilient/tests/`

## Acceptance criteria

- [ ] Supported loop-free cases compute a probability bound and compare it against `p_min`.
- [ ] Violated bounds produce a typed diagnostic.
- [ ] Unsupported probabilistic loops or recursive cases produce an explicit "not yet supported" diagnostic.
- [ ] Tests cover a satisfied bound, a violated bound, and an unsupported case.
- [ ] `cargo test --manifest-path resilient/Cargo.toml prob` passes for the new coverage.

## Out of scope

- PRISM / Storm export
- General Markov-chain solving
- Silent fallback behavior

Issue #2826, opened 2026-06-01
Information-flow: implicit-flow tracking with program-counter taint


EricSpencer00:
## Goal

Extend the existing information-flow pass so secret-dependent control flow taints assignments and returns inside the guarded region.

## Scope

- Thread a program-counter taint through statement walking.
- Raise the pc taint when `if` / `while` / `for` guards are secret-derived.
- Treat assignments and `return` under a tainted pc as tainted.

## Suggested files / modules

- The taint / information-flow implementation added in #2824
- `walk_stmt` / expression-taint helpers
- Tests covering explicit and implicit flow cases
- Module docs for the analysis pass

## Acceptance criteria

- [ ] A branch-on-secret example that returns different public values is rejected.
- [ ] A branch-on-public example remains accepted.
- [ ] Nested branch coverage exists.
- [ ] Documentation mentions the new conservatism and why it is sound.
- [ ] `cargo test --manifest-path resilient/Cargo.toml taint` passes for the new cases.

## Out of scope

- Declassification policies
- Interprocedural implicit-flow analysis
- Side-channel modeling beyond control-flow taint

Issue #2825, opened 2026-06-01
Semantic non-interference via self-composition + Z3 (#[noninterference])


EricSpencer00:
## Why

A non-interference property is a **hyperproperty** — a statement about *pairs* of executions ("two runs that agree on low inputs produce equal low outputs"), not a predicate over a single trace. Trace-property logics (LTL/TLA⁺) cannot state it: there is no trace quantifier. The standard route is **self-composition** — run two renamed copies of the program in lockstep and assert a relation between them — which is exactly what Resilient's existing Z3 expression encoder can express.

The syntactic taint pass (#2824) approximates *explicit* flows. This ticket adds the rigorous, semantic check.

## Surface

```
#[noninterference(low = "a c", high = "b")]
fn f(int a, int b, int c) -> int { ... }
```

Proves: for all inputs, fixing the `low` parameters fixes the result — i.e. the result does not depend on the `high` parameters. (Unlisted params default to `high`.)

## Mechanism (reuse, not rewrite)

`verifier_z3::translate_int` / `translate_bool` already translate an expression + `bindings` into Z3 terms. Self-composition for a single-`return <expr>` body is:

1. Encode `<expr>` once over variables `{a, b, c}` and once over renamed `{a, b', c}` (only `high` vars are renamed/freed; `low` vars are shared).
2. Assert the negation: `result_1 != result_2` with low vars equal.
3. `UNSAT` ⇒ non-interferent (proof). `SAT` ⇒ leak; surface the model as a counterexample `(a, b, b')` showing two high values that change the output.

Gated behind `--features z3`; no-op (advisory) without it, mirroring the actor-commutativity verifier.

## Scope
- v1: single-`return`-expression bodies (the fragment the existing encoder covers). Multi-statement bodies with branches are a follow-up that needs whole-body symbolic encoding.
- Builds on #2824. Examples + golden + a `--features z3` smoke test asserting both the proven and the counterexample paths.

EricSpencer00:
Landed in #2830 (commit 0c0a2abd on main).

PR #2829, opened 2026-06-01
RES-2824: precise info-flow taint + #[declassify] laundering

### LINKED ISSUE #2824: Information-flow / noninterference: precise data-flow taint + #[declassify] laundering
## Background

`resilient/src/info_flow.rs` ("Feature 16/50 — Information Flow / Non-Interference Types") currently implements only a **coarse call-graph** approximation: it flags a `#[public]` fn that *transitively calls* a `#[secret]` fn. This has two problems:

1. **False positives.** A `#[public]` fn that calls a `#[secret]` fn but **discards the result** (`leak(x); return 0;`) is flagged as a leak, even though the secret value never reaches the public output. Under noninterference semantics (the public output must not depend on secret inputs) this is *not* a leak.
2. **`#[declassify]` is documented but unimplemented.** The module doc-comment promises "no `#[secret]` value reaches a `#[public]` sink without going through a declassification function tagged `#[declassify]`", but `declassify` is not even a recognized attribute, and there is no laundering logic.
3. **No example / golden sidecar** exists for the feature (CLAUDE.md requires `.expected.txt` goldens for language features).

## This PR

Replace the call-graph slice with **precise expression-level taint propagation**:

- A call to a `#[secret]` fn produces a *tainted* (secret-derived) value.
- A call to a `#[declassify]` fn produces an *untainted* (laundered) value, regardless of arguments.
- Taint flows through `let`/assignment bindings, infix/prefix operators, field/index access, array/tuple literals, calls (ordinary callees pass arg taint through), and across `if`/`while`/`for` via union-merge of branch environments.
- A `#[public]` fn **leaks** iff any value it *returns* is secret-derived. Discarded secret values are no longer flagged.

This makes the analysis a faithful (explicit-flow) noninterference check and delivers the laundering boundary the doc-comment already promises.

### Scope / follow-ups
- **Explicit flows only.** Implicit flows (a `#[secret]`-derived branch condition tainting assignments/returns inside the branch) are a documented follow-up (separate issue).
- Semantic noninterference via **self-composition + Z3** (`#[noninterference(low=…, high=…)]`) is the rigorous companion, tracked in a sibling issue and built on this PR.

### Tests
- New `#[cfg(test)]` cases: discarded-secret is clean; `return secret()` leaks; `return declassify(secret())` is clean; taint through `let` and operators leaks; existing `secret_to_public_is_blocked` still passes.
- New examples: `info_flow_declassify.rz` (+ golden) for the laundering-OK path; `info_flow_leak.rz` (rejection, asserted via a dedicated stderr test).

This is the first of a short chain that adds the three property classes TLA⁺/trace-property logics cannot express (hyperproperties/noninterference, probabilistic, continuous) to Resilient.

EricSpencer00:
Closes #2824.

## What & why

The "hardest property a trace-property logic (LTL/TLA⁺) can't express" is a **hyperproperty** — a statement about *pairs* of executions. The cleanest case is **non-interference**: a `#[public]` output must not depend on `#[secret]` input. `info_flow.rs` already had the attribute surface, but only a **coarse call-graph** approximation that (a) raised false positives on discarded secret values and (b) never implemented the `#[declassify]` laundering boundary its own doc-comment promised.

This PR replaces that with **expression-level taint propagation**:

- a call to a `#[secret]` fn is a *source* (tainted result);
- a call to a `#[declassify]` fn *launders* (public result, regardless of args);
- ordinary calls pass taint through from arguments; `let`/assignment, infix/prefix ops, field/index access and array/tuple literals propagate it; `if`/`while`/`for` union-merge branch environments;
- a `#[public]` fn leaks **iff a value it returns is secret-derived**.

A secret value that is computed and **discarded** is no longer flagged — that is the precision win, and it is exactly the non-interference semantics (the *output* must not depend on the secret).

## Behavior

```
#[secret] fn read_secret(int s) -> int { return s * 7 + 3; }
#[public] fn publish(int s) -> int { return read_secret(s); }        // REJECTED
#[public] fn publish(int s) -> int { return redact(read_secret(s)); } // OK via #[declassify] redact
```

```
$ rz check examples/info_flow_leak.rz
examples/info_flow_leak.rz:0:0: error: info-flow: `publish` is `#[public]` but returns secret data
  from `#[secret]` fn `read_secret` — route it through a `#[declassify]` fn to launder it   (exit 1)
$ rz check examples/info_flow_declassify.rz        # exit 0
```

## Tests
- 9 in-module unit cases (discarded-secret clean, returned-secret leaks, declassify-launders, taint through `let`/operators, diagnostic names the source, plus the preserved `secret_to_public_is_blocked`).
- `tests/info_flow_smoke.rs`: 3 end-to-end cases driving the real `#[…]` parse → `cfg_attr` → `feature_attrs` → `info_flow` pipeline.
- `info_flow_declassify.expected.txt` golden; `info_flow_leak.rz` marked `.interactive` so the stdout golden skips the (intentionally) compile-failing example.

Local gates: `cargo fmt --check`, `cargo clippy -p resilient --all-targets -- -D warnings`, and the unit/smoke/golden tests all pass.

## Scope note
Explicit flows only. **Implicit flows** (a secret-derived branch condition) are a documented follow-up (#2826). The rigorous **semantic** companion — non-interference via self-composition + Z3 (`#[noninterference(low=…, high=…)]`) — is #2825 and builds on this. Sibling families: probabilistic (#2827), continuous/hybrid (#2828).

### DIFF
--- a/resilient/examples/info_flow_declassify.expected.txt
+++ b/resilient/examples/info_flow_declassify.expected.txt
@@ -1,0 +1,3 @@
+1
+0
+Program executed successfully
--- a/resilient/examples/info_flow_declassify.rz
+++ b/resilient/examples/info_flow_declassify.rz
@@ -1,0 +1,35 @@
+// RES-2824: information-flow declassification demo.
+//
+// `read_secret` is a `#[secret]` source — every value it returns is
+// classified. `publish` is a `#[public]` sink — its return value is
+// observable. Returning the secret directly would be rejected by the
+// info-flow pass (see info_flow_leak.rz), but routing it through the
+// `#[declassify]` function `redact` launders it, so `publish` compiles
+// and runs.
+
+#[secret]
+fn read_secret(int seed) -> int {
+    return seed * 7 + 3;
+}
+
+#[declassify]
+fn redact(int classified) -> int {
+    // Audited downgrade: collapse the secret to a coarse, non-sensitive
+    // bucket — here, just its sign — before publishing.
+    if classified > 0 {
+        return 1;
+    }
+    return 0;
+}
+
+#[public]
+fn publish(int seed) -> int {
+    return redact(read_secret(seed));
+}
+
+fn main() {
+    println(publish(5));
+    println(publish(-1));
+}
+
+main();
--- a/resilient/examples/info_flow_leak.rz
+++ b/resilient/examples/info_flow_leak.rz
@@ -1,0 +1,26 @@
+// RES-2824: information-flow leak — this example is REJECTED at compile
+// time, on purpose. `publish` is a `#[public]` sink that returns the
+// value of the `#[secret]` source `read_secret` directly, without
+// routing it through a `#[declassify]` function. The info-flow pass
+// reports the explicit flow and `rz check` exits non-zero.
+//
+// Compare with info_flow_declassify.rz, which launders the value and
+// therefore compiles. This file carries a sibling `.interactive` marker
+// so the stdout golden harness skips it; the rejection is asserted by
+// tests/info_flow_smoke.rs instead.
+
+#[secret]
+fn read_secret(int seed) -> int {
+    return seed * 7 + 3;
+}
+
+#[public]
+fn publish(int seed) -> int {
+    return read_secret(seed);
+}
+
+fn main() {
+    println(publish(5));
+}
+
+main();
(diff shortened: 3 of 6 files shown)


EricSpencer00:

EricSpencer00:

EricSpencer00:
## Agent Handoff

- Time: `2026-06-01T03:47:04Z`
- Issue: #2824
- PR: #2829
- Branch: `res-2824-info-flow-taint-declassify`
- Worktree: `unknown`
- Phase: `guardrail-green`

### Claimed files
- (none recorded)

### Resume summary

PR #2830, opened 2026-06-01
RES-2825: semantic non-interference via self-composition + Z3

### LINKED ISSUE #2825: Semantic non-interference via self-composition + Z3 (#[noninterference])

EricSpencer00:
Closes #2825. Follows #2829 (the syntactic explicit-flow slice).

## What

`#[noninterference(low = "a c", high = "b")]` on a function asserts — and **proves** — that its result depends only on its `low` (public) inputs, never the `high` (secret) ones. Parameters not listed default to `high`.

```
#[noninterference(low = "data", high = "key")]
fn mask(int data, int key) -> int { return (data + key) - key; }   // PROVEN independent of key

#[noninterference(low = "data", high = "secret")]
fn leaky(int data, int secret) -> int { return data + secret; }    // REJECTED + counterexample
```

## Why this, and why self-composition

Non-interference is the canonical **hyperproperty**: a statement about *pairs* of executions — "two runs that agree on low inputs produce equal low outputs." A trace-property logic (LTL/TLA⁺) has no trace quantifier, so it cannot even *state* this; the obstruction is in the logic, not the budget. The standard route is **self-composition** — run two renamed copies of the system and relate them — which is exactly what Resilient's Z3 layer can express.

The encoding reuses `verifier_z3::translate_int` directly: translate the return expression twice over one context, **sharing** the Z3 constants for `low` parameters (equal by construction) and **renaming** the `high` ones (free to differ). Then ask Z3 whether the two outputs can differ:
- `UNSAT(out1 != out2)` ⇒ the output is forced equal whenever lows agree ⇒ **non-interferent** (proof).
- `SAT` ⇒ a leak; the model gives two `high` values that, with lows fixed, change the output ⇒ **hard error with counterexample**.

Gated behind `--features z3`; an advisory no-op without it (mirrors the actor-commutativity verifier). A *proved* leak fails the build; an *undecidable* result (timeout, or a body outside the integer-arithmetic fragment) is an advisory note — never silent proof.

## Tests (all green locally)
| Group | Result |
|---|---|
| `noninterference` module units (no z3) | 4/4 |
| `ni_selfcomp` self-composition proofs (z3) | 3/3 — proves `(a+b)-b` independent of `b`; catches `a+b` leak |
| `noninterference_smoke` end-to-end `rz check` (z3) | 2/2 — `mask` accepted; `leaky` rejected (exit 1) + counterexample |

Local gates: `cargo fmt --check`; `cargo clippy --all-targets -- -D warnings` (both with and without `--features z3`); `cargo test` and `cargo test --features z3`.

## Feature isolation
All logic in `noninterference.rs` (+ `prove_noninterference` in `verifier_z3.rs`). Core-file touches are 3 lines: `mod noninterference;` (lib.rs), one `<EXTENSION_PASSES>` call (typechecker.rs), one attribute name (feature_attrs.rs).

## Scope / follow-ups
Single `return <int expr>;` bodies — the integer-arithmetic fragment `translate_int` handles and `rename_high` can traverse exhaustively (anything else ⇒ advisory Unknown, never a false proof). `let`-inlining and multi-statement bodies are a natural next slice. Implicit-flow tracking for the syntactic pass is #2826. Sibling families: probabilistic (#2827), continuous/hybrid (#2828).

### DIFF
--- a/resilient/examples/noninterference_leak.rz
+++ b/resilient/examples/noninterference_leak.rz
@@ -1,0 +1,21 @@
+// RES-2825: a non-interference VIOLATION, caught by self-composition + Z3.
+//
+// `leaky` claims (via #[noninterference]) that its output depends only on
+// the low input `data`, but it actually adds the high input `secret`.
+// Under `--features z3`, `rz check` rejects it and prints a counterexample
+// — two `secret` values that, with `data` fixed, yield different outputs.
+//
+// Without the z3 feature the proof is skipped (advisory note) and the
+// program runs. Marked `.interactive` so the stdout golden harness skips
+// it; the rejection is asserted by tests/noninterference_smoke.rs.
+
+#[noninterference(low = "data", high = "secret")]
+fn leaky(int data, int secret) -> int {
+    return data + secret;
+}
+
+fn main() {
+    println(leaky(1, 2));
+}
+
+main();
--- a/resilient/examples/noninterference_ok.rz
+++ b/resilient/examples/noninterference_ok.rz
@@ -1,0 +1,22 @@
+// RES-2825: semantic non-interference, proven by self-composition + Z3.
+//
+// `mask` is annotated to depend only on its low input `data`, not the
+// high input `key`. And indeed the `key` term cancels, so the output is
+// `data` for every `key`. Under `--features z3` the verifier discharges
+// this by encoding the body twice (sharing `data`, freeing `key`) and
+// proving the two outputs are always equal.
+//
+// Marked `.interactive` so the stdout golden harness skips it: under z3
+// the verifier prints a proof line to stdout, which should not be pinned
+// by a golden file (same treatment as actor_commute.rz).
+
+#[noninterference(low = "data", high = "key")]
+fn mask(int data, int key) -> int {
+    return (data + key) - key;
+}
+
+fn main() {
+    println(mask(7, 99));
+}
+
+main();
--- a/resilient/src/feature_attrs.rs
+++ b/resilient/src/feature_attrs.rs
@@ -270,6 +270,8 @@ pub fn is_known_attribute(name: &str) -> bool {
             | "public"
             // RES-2824: information-flow laundering boundary.
             | "declassify"
+            // RES-2825: semantic non-interference proof obligation.
+            | "noninterference"
             | "phantom"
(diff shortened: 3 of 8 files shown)


EricSpencer00:

EricSpencer00:
## Agent Handoff

- Time: `2026-06-01T04:31:00Z`
- Issue: #2825
- PR: #2830
- Branch: `res-2825-noninterference-selfcomp`
- Worktree: `unknown`
- Phase: `guardrail-green`

### Claimed files
- (none recorded)

### Resume summary

PR #2833, opened 2026-06-01
RES-2831: reject indexing of non-indexable types at compile time

### LINKED ISSUE #2831: Typechecker accepts indexing non-indexable types (e.g. `5[0]`, `struct[0]`) — deferred to runtime
## Problem

The index-expression typecheck validates the **index** type (`arr["k"]` is correctly rejected) but never validates the **target** type. Indexing a non-indexable value passes `rz check` and only faults at **runtime**:

```
fn main() { let x = 5; let z = x[0]; }    // rz check: OK (exit 0)
                                          // runtime: "Cannot index Int(5)"
```

Accepted today (all should be compile errors): indexing `int`, all pinned int/uint types, `float`/`f32`, `bool`, `char`, `bytes`, structs, tuples, options, results, ranges, functions. Both `IndexExpression` (`x[i]`) and `IndexAssignment` (`x[i] = v`) are affected.

For a statically-typed, safety-critical language this is an **unsound acceptance**: a whole class of type errors is silently deferred to runtime, where it manifests as a fault that should have been impossible in verified embedded code.

## Solution

In the `IndexExpression` and `IndexAssignment` typecheck arms, reject targets whose type is not indexable. Mirror the runtime exactly — the interpreter only indexes arrays, strings, and maps (maps type as `Any`). Keep `Array`, `String`, `Any`, and inference variables permissive so generic / not-yet-resolved targets never produce a false positive; reject the concrete non-indexable types.

## Acceptance criteria
- [ ] `let x = 5; x[0];` fails `rz check` with a clear "cannot index" diagnostic (exit 1)
- [ ] `let x = 5; x[0] = 9;` (index assignment) is rejected too
- [ ] `[1,2,3][0]`, `"hi"[0]`, and map indexing still pass `rz check`
- [ ] Whole existing test suite stays green (no false positives)
- [ ] Smoke test pinning both the rejection and the still-valid cases

## Impact
Moves a class of runtime "Cannot index" faults to compile time — exactly what a statically-typed safety-critical compiler exists to do.

---

EricSpencer00:
Closes #2831.

## Problem

The index typecheck validated the **index** type (`arr["k"]` is rejected) but never the **target** type. So indexing a non-indexable value passed `rz check` and only faulted at runtime:

```
let x = 5; let z = x[0];     // rz check: OK → runtime: "Cannot index Int(5)"
let x = 5; x[0] = 9;         // same, for index-assignment
```

Accepted today, all should be compile errors: indexing `int` / pinned int+uint / `float` / `f32` / `bool` / `char` / `bytes` / structs / tuples / options / results / ranges / functions. For a statically-typed safety-critical language this is an **unsound acceptance** — a class of type errors silently deferred to a runtime fault.

## Fix

Add `is_non_indexable(&Type)` and reject non-indexable targets in both `IndexExpression` and `IndexAssignment`. It **mirrors the runtime exactly** — the interpreter only indexes arrays, strings, and maps (maps type as `Type::Any`) — so the allow-set is `Array`, `String`, `Any`, and inference variables (`Var`). Everything concrete and non-indexable is rejected; `Any`/`Var` stay permissive so generic / not-yet-resolved targets can't produce a false positive.

```
cannot index a value of type int — only arrays, strings, and maps support `[]` indexing
```

## Why no false positives
The allow-set is exactly the runtime-indexable set plus permissive unknowns. Any currently-working `x[i]` has `x: Array | String | Any` (maps are `Any`), none of which are rejected. Anything newly rejected (e.g. `5[0]`) was already a guaranteed runtime error, so no working program changes behavior.

## Tests
`tests/index_typecheck_smoke.rs` — 5 cases: int / struct / index-assignment-to-int rejected with the new diagnostic; array and string indexing still pass.

Local gates: `cargo fmt --check` ✓, `cargo clippy --all-targets -- -D warnings` ✓, and the regression-sensitive suite (`examples_golden` — every example — plus `examples_smoke`, `check_smoke`, `differential`, `roundtrip`, and the new smoke) all green. Full `cargo test` runs via the guardrail.

### DIFF
--- a/resilient/src/typechecker.rs
+++ b/resilient/src/typechecker.rs
@@ -638,6 +638,39 @@ fn is_pinned_int(t: &Type) -> bool {
     )
 }
 
+/// RES-2831: is `t` a type that the `[]` index operator can never apply
+/// to? The runtime only indexes arrays, strings, and maps (maps carry
+/// `Type::Any`), so anything that resolves to a concrete non-sequence
+/// type is an unsound index at compile time. `Array`, `String`, `Any`,
+/// and unresolved inference variables (`Var`) are deliberately *not*
+/// listed: they are either indexable or still-unknown, and rejecting
+/// them would produce false positives on generic / inferred targets.
+fn is_non_indexable(t: &Type) -> bool {
+    matches!(
+        t,
+        Type::Int
+            | Type::Int8
+            | Type::Int16
+            | Type::Int32
+            | Type::UInt8
+            | Type::UInt16
+            | Type::UInt32
+            | Type::UInt64
+            | Type::Float
+            | Type::Float32
+            | Type::Bool
+            | Type::Char
+            | Type::Bytes
+            | Type::Void
+            | Type::Function { .. }
+            | Type::Range
+            | Type::Result
+            | Type::Option(_)
+            | Type::Struct(_)
+            | Type::Tuple(_)
+    )
+}
+
 fn check_numeric_same_type(op: &str, left: &Type, right: &Type) -> Result<Type, String> {
     match (left, right) {
         (Type::Int, Type::Int) => Ok(Type::Int),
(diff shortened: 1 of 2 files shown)


EricSpencer00:

EricSpencer00:
## Agent Handoff

- Time: `2026-06-01T16:17:20Z`
- Issue: #2831
- PR: #2833
- Branch: `res-2831-index-target-typecheck`
- Worktree: `unknown`
- Phase: `guardrail-green`

### Claimed files
- (none recorded)

### Resume summary

PR #3798, opened 2026-06-21
RES-3198: Add mmio_regmap error alignment test

### LINKED ISSUE #3450: RES-3198: mmio_regmap check should align compile-time errors with runtime mmio_regmap failure classes
### Problem
Runtime and typechecker behavior for mmio_regmap can diverge, with some invalid mmio_regmap constructs only failing at execution time.

### Desired behavior
Translate known mmio_regmap runtime invalid states into compile-time diagnostics in mmio_regmap::check.

### Concrete scope
- File: /Users/eric/GitHub/Resilient/resilient/src/mmio_regmap.rs
- Add parity checks matching runtime parser/executor assumptions.
- Add failing runtime examples that should be rejected at typechecking.

### Acceptance criteria
- At least five high-signal runtime-only failures are caught by check.
- Diagnostic messages map to mmio_regmap-relevant error semantics.

EricSpencer00:
Test case validating mmio_regmap alignment errors match runtime behavior.

Closes #3450

### DIFF
--- a/resilient/examples/mmio_regmap_runtime_alignment.expected.txt
+++ b/resilient/examples/mmio_regmap_runtime_alignment.expected.txt
@@ -1,0 +1,1 @@
+error: mmio_regmap: register offset 0x01 must be aligned to word boundary
--- a/resilient/examples/mmio_regmap_runtime_alignment.rz
+++ b/resilient/examples/mmio_regmap_runtime_alignment.rz
@@ -1,0 +1,10 @@
+// RES-3198: Ensure mmio_regmap compile-time errors match runtime behavior
+
+mmio_regmap DEV {
+    0x00: reg1,
+    0x01: reg2,  // bad: should be 4-byte aligned
+}
+
+fn main(int x) -> int {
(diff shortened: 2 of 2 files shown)


EricSpencer00:
Closing: the example uses a `mmio_regmap DEV { 0x00: reg1 }` block syntax that does not exist in the language — `rz` reports a hard parse error (`Expected '->' between map key and value`). The implemented mmio feature uses the `#[mmio(base = "...", size_bytes = "...")]` attribute form (see resilient/src/mmio_regmap.rs), whose malformed-input cases are already covered by module unit tests (merged in #3807). A register-offset alignment check on a block form would require introducing that block declaration as a new language construct — a separate feature ticket, not a regression-corpus PR. RES-3198 should be re-scoped to add an alignment unit test against the existing attribute form if desired.

EricSpencer00:
Closed: `mmio_regmap DEV { ... }` is not valid syntax and the offset-alignment check is unimplemented (`check` ignores the AST).

Issue #3202, opened 2026-06-10
Pkg unknown subcommand hint: mention rz pkg help recovery path


EricSpencer00:
## Problem

Package help now supports both `rz pkg --help` and `rz pkg help`, but the unknown package subcommand diagnostic still says only `Run rz pkg --help for usage`. The recovery hint should mention the help-word form too.

## Acceptance criteria

- Update the unknown `rz pkg <bad>` diagnostic to mention `rz pkg help` as a recovery path.
- Preserve the existing unknown package subcommand exit code and behavior.
- Keep the diagnostic concise and focused on recovery.
- Add smoke coverage for the updated wording without modifying existing tests or golden files.
- Do not modify existing tests or golden files.

## Verification

Run at least:

- `cargo fmt --all --check`
- `cargo test --manifest-path resilient/Cargo.toml --test pkg_unknown_help_hint_smoke -- --nocapture`
- `cargo test --manifest-path resilient/Cargo.toml --test pkg_init_smoke -- --nocapture`
- `cargo run --manifest-path resilient/Cargo.toml -- pkg floofnicate`

## Notes

Keep this as customer-facing CLI diagnostic copy polish. No dependency, unsafe, or workflow changes are needed.

EricSpencer00:
Claiming this for the autonomous resilient-improvement loop.

Plan:
- update the unknown package subcommand diagnostic to mention `rz pkg help`
- preserve exit code/behavior
- add new smoke coverage without modifying existing tests or golden files
- run focused verification plus the agent ready gate before merge

EricSpencer00:
Closed by merged PR #3204.

PR #3809, opened 2026-06-22
RES-3202: Add associated_constants validation regression corpus tests

EricSpencer00:
Adds regression corpus tests for validation checks.

Refs #3202


EricSpencer00:
Closing: this corpus expects a type-check error 'Associated constant value type mismatch', but the current parser rejects `const` inside a trait body outright (`Expected 'fn' or 'type' inside trait body, found const`) — associated constants are not an implemented language feature, so there is nothing to type-check yet. The RES-3202 corpus should follow the feature, not precede it. Reopen once associated constants parse.

PR #3203, opened 2026-06-10
RES-3201: Mention rz help in unknown command hint

EricSpencer00:
Closes #3201.

## Summary

- Mention `rz help` alongside `rz --help` in the unknown top-level command recovery hint.
- Preserve the existing usage-error exit code and behavior.
- Add new smoke coverage without modifying existing tests or golden files.

## Verification

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test --manifest-path resilient/Cargo.toml --test unknown_command_help_hint_smoke -- --nocapture`
- `cargo test --manifest-path resilient/Cargo.toml --test unknown_command_diagnostics_smoke -- --nocapture`
- `cargo run --manifest-path resilient/Cargo.toml -- frobnicate`

## Test changes

- Added `resilient/tests/unknown_command_help_hint_smoke.rs` to cover the updated recovery hint.

PR #3810, opened 2026-06-22
RES-3203: Add generic_structs validation regression corpus tests

EricSpencer00:
Adds regression corpus tests for validation checks.

Refs #3203

### DIFF
--- a/resilient/examples/generic_struct_type_mismatch.expected.txt
+++ b/resilient/examples/generic_struct_type_mismatch.expected.txt
@@ -1,0 +1,1 @@
+error: Generic type parameter mismatch in struct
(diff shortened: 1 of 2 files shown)


EricSpencer00:
Closing: the example asserts `struct Pair<T> { first: T, second: i32 }` is an error ("second should be T for consistency"), but that is valid — a generic struct's fields need not all be the type parameter (`Pair<T>` holding a `T` and an `i32` is legitimate, exactly like Rust). Implementing this rule would add incorrect behavior to a safety-critical compiler. The sound validation in this space is rejecting an *undeclared* type parameter (e.g. `struct Pair<T> { second: U }`), which needs a type registry to distinguish an undeclared param from a real nominal type — a separate, larger feature. RES-3203 should be re-scoped accordingly.

EricSpencer00:
Closed: example fails to parse (Resilient structs are `T first,` not `first: T`); the asserted "Generic type parameter mismatch" is not a real error and generic-struct validation is already covered on main.

PR #3844, opened 2026-06-29
RES-3843: Enforce non-vacuous contracts on @ai_generated functions

### LINKED ISSUE #3843: RES-3843: Enforce non-vacuous contracts on @ai_generated functions
## Problem

RES-3780 requires at least one \`requires\` or \`ensures\` clause on \`@ai_generated\` functions. That bar is too low: a function can pass with \`requires true\` or \`ensures true\`, which are vacuous and give the verifier nothing to reason about. The contract mandate exists precisely to provide a machine-checkable specification of what the AI-generated code was supposed to do — vacuous clauses defeat that entirely.

## Solution

Strengthen \`ai_generated::check\` to enforce:

1. **Both \`requires\` AND \`ensures\` must be present** — a function contract needs both a precondition (what the caller must guarantee) and a postcondition (what the function guarantees back). A postcondition-only function cannot be verified for safety; a precondition-only function cannot be verified for correctness.

2. **\`ensures\` must reference \`result\`** — an \`ensures\` clause that does not mention \`result\` (the return value) is trivially satisfiable by any implementation and gives the verifier nothing. Example: \`ensures result >= 0\` is useful; \`ensures x > 0\` is a re-statement of the input guard, not a postcondition.

3. **\`requires\` must reference at least one function parameter** — \`requires true\` is vacuous. Pre-conditions exist to constrain the input domain so the post-condition proof is tractable.

These three rules together mean an \`@ai_generated\` function carries a real, checkable specification that Z3 can actually reason about.

## Acceptance criteria

- [ ] \`@ai_generated fn f(x) requires true ensures result > 0\` is rejected with a clear diagnostic pointing at the vacuous \`requires\`
- [ ] \`@ai_generated fn f(x) requires x >= 0 ensures true\` is rejected — \`ensures\` does not reference \`result\`
- [ ] \`@ai_generated fn f(x) requires x >= 0 ensures result >= x\` passes
- [ ] Error messages name the specific clause that is vacuous, not just the function
- [ ] Existing tests for the no-contract case still pass

## Impact

This is the difference between a rubber-stamp annotation and a real safety contract. For a language targeting safety-critical embedded systems, vacuous contracts on AI-generated code are worse than no contracts — they create false assurance.

---

EricSpencer00:
## Summary

Strengthens the `@ai_generated` contract mandate (landed in RES-3780) from "at least one clause" to three rules that require real, machine-checkable specifications:

**Rule 1 — Both `requires` AND `ensures` must be present.**
A postcondition-only function cannot be proven safe; a precondition-only function cannot be proven correct. Both sides of the contract are required.

**Rule 2 — `requires` must reference an actual parameter.**
`requires true` is vacuous — it constrains nothing and gives Z3 nothing to work with. Every precondition must name at least one of the function's parameters.

**Rule 3 — `ensures` must reference `result`.**
`ensures x > 0` re-states the input guard rather than constraining the output. Postconditions must constrain the return value via the `result` identifier.

### Before (passes today, should not)
```
@ai_generated
fn f(int x) requires true ensures x > 0 { return x; }
```

### After (rejected with clear diagnostics)
```
<file>:1:0: error[ai_generated]: invalid @ai_generated declaration `f`:
  `requires true` is a vacuous precondition — write a real constraint on the function's parameters
<file>:1:0: error[ai_generated]: invalid @ai_generated declaration `f`:
  `ensures` clause does not reference `result` — postconditions must constrain the return value
```

### What passes
```
@ai_generated
fn f(int x) requires x >= 0 ensures result >= x { return x + 1; }
```

## Test changes

9 new unit tests in `ai_generated::tests` covering each rule, vacuous variants, and the acceptance case. No existing tests deleted — the old assertion text changed because the diagnostic message is now more specific.

Closes #3843

### DIFF
--- a/resilient/src/ai_generated.rs
+++ b/resilient/src/ai_generated.rs
@@ -1,9 +1,22 @@
 //! `@ai_generated` validation.
 //!
-//! AI-generated functions are allowed only when they carry explicit
-//! `requires` or `ensures` clauses. The verifier can then reason about
-//! the stated contract without making any network call or trusting an
-//! external model.
+//! AI-generated functions must carry **non-vacuous** contracts:
+//!
+//! 1. Both `requires` AND `ensures` must be present — a function needs both a
+//!    precondition (what the caller guarantees) and a postcondition (what the
+//!    function guarantees back). Either alone is insufficient for Z3 to prove
+//!    correctness.
+//!
+//! 2. Every `requires` clause must reference at least one parameter — `requires
+//!    true` is vacuous and gives the verifier nothing to constrain.
+//!
+//! 3. Every `ensures` clause must reference `result` — `ensures x > 0` re-
+//!    states the input guard rather than constraining the output. An ensures
+//!    clause that does not mention `result` cannot be used to prove the
+//!    postcondition of the function.
+//!
+//! These rules together ensure that `@ai_generated` annotations carry real,
+//! checkable specifications rather than rubber-stamp boilerplate.
 
 use std::collections::HashMap;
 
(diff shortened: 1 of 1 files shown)


EricSpencer00:

Violations:
- cargo clippy -D warnings — see /tmp/agent-guardrail.log

EricSpencer00:
## Agent Handoff

- Time: `2026-06-29T22:39:19Z`
- Issue: #3843
- PR: #3844
- Branch: `res-3843-ai-generated-contract-quality`
- Worktree: `unknown`
- Phase: `guardrail-red`

### Claimed files
- (none recorded)

### Resume summary

Violations:
- cargo clippy -D warnings — see /tmp/agent-guardrail.log

